PR #4, opened 2024-01-28
Added trunk linting and a few new sample source apps

jwmatthews:
This adds https://docs.trunk.io/check#install-the-cli to be used for linting/formatting


savitharaghunathan (approved):
/lgtm

PR #8, opened 2024-01-30
Loading Incident store

savitharaghunathan:
My first stab at incident store

1. There is a load method that takes the list of applications and loads the cached violations data store.
Sample command to run `./run_cli.py load kitchensink,eap-coolstore-monolith`

2. writes to `samples/generated_output/incident_store/cached_violations.yaml`

### DIFF
--- a/kai/incident_store.py
+++ b/kai/incident_store.py
@@ -84,6 +84,7 @@ class IncidentStore:
                        file_path: [ {
                             variables: {}
                             line_number: int
+                            message: str
                         } ]
                     }
                }
--- a/kai/incident_store.py
+++ b/kai/incident_store.py
@@ -1,0 +1,292 @@
+__all__ = ["Application", "IncidentStore"]
+
+import copy
+import os
+import pprint
+
+import yaml
+
+from kai.report import Report
+
+
+class Application:
+    def __init__(
+        self,
+        name,
+        report,
+        repo=None,
+        initial_branch=None,
+        solved_branch=None,
+        commitId=None,
+        timestamp=None,
+    ):
+        self.name = name
+        self.report = report
+        self.repo = repo
+        self.initial_branch = initial_branch
+        self.solved_branch = solved_branch
+        self.commitId = commitId
+        self.timestamp = timestamp
+
+
+class IncidentStore:
+    def __init__(self):
+        # applications, keyed on App Name
+        self.applications = {}
+        # cached_violations is defined in comments in self._update_cached_violations
+        self.cached_violations = {}
+        self.pp = pprint.PrettyPrinter(indent=2)
+
+    def add_app_to_incident_store(self, app_name, yaml):
+        r = Report(yaml).get_report()
+        a = Application(name=app_name, report=r)
+        self.applications[app_name] = a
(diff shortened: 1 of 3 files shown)


kai/incident_store.py
```diff
@@ -0,0 +1,292 @@
...
+
+from kai.report import Report
+
+
+class Application:
+    def __init__(
+        self,
+        name,
+        report,
+        repo=None,
```
jwmatthews, line 17: In future we will probably need to update 'repo' so that we have flexibility for case where repo-A is the original and repo-B is the updated version.  Less of a concern when integrated in Konveyor, more that this flexibility will help us earlier with testing samples we collect.

jwmatthews: We don't need to update this in this PR, just sharing this note for thinking through/discussing.  We can adjust it later when it's obvious we need it.

kai/incident_store.py
```diff
@@ -0,0 +1,292 @@
...
+    def _update_cached_violations(self, a):
+        """
+        Update the cached_violations with the new application
+        The desired structure
+         cached_violations: {
+           ruleset_name: {
+               violation_name: {
+                   app_name: {
+                       file_path: [ {
+                            variables: {}
```
jwmatthews, line 85: minor 'ni':  May be worth calling out we have 'message' in this as well for future folks who come back to look at comment.

jwmatthews (approved):
LGTM
This matches what I was thinking for starting out, looks good @savitharaghunathan

kai/incident_store.py
savitharaghunathan: I went through a few samples and message field is a part of variables. By calling out, do we want to bring it out of variables and have separate field or just a mechanism to make sure that a 'message' is available for this incident. When I say mechasim, I am thinking of a simple helper function that can quickly lookup the incident and tell me if there is a message or not.

IMO, the message field variables will be very useful in building prompts.

jwmatthews: I made an assumption that this data structure mapped to what you generate when you run ` ./run_cli.py load kitchensink,eap-coolstore-monolith` and write to samples/incident_datastore

Maybe a bad assumption on my part.  I saw how you had 'message' in the cached data and thought that was helpful.  When I looked at it, it seemed like message was on same level as 'line_number' and 'variables', so I thought it would help to update this comment for future folks looking through code.


```
      src/main/resources/META-INF/resources/bower_components/angular-patternfly/dist/docs/img/RH_atomic.svg:
      - line_number: 4
        message: '`beans_1_1.xsd`: In the root tag, replace the `version` attribute
          value with `3.0`'
        variables:
          matchingText: version="1.1"
      src/main/resources/META-INF/resources/bower_components/angular-patternfly/dist/docs/img/bg-navbar-pf-alt.svg:
      - line_number: 4
        message: '`beans_1_1.xsd`: In the root tag, replace the `version` attribute
          value with `3.0`'
        variables:
          matchingText: version="1.1"
```

savitharaghunathan: ACK. Got it. Thanks for the clarification. I will update the comment

PR #9, opened 2024-01-31
Adding trunk check to run on PRs

jwmatthews:
Following guidance from https://github.com/trunk-io/trunk-action?tab=readme-ov-file#getting-inline-annotations-for-fork-prs


jmontleon (approved):
ACK

PR #20, opened 2024-02-02
More lint fixes

jmontleon:


.trunk/configs/.bandit
```diff
@@ -0,0 +1,2 @@
+[bandit]
+skips = B101
```
jmontleon, line 2: This check is nonsensical for tests.
https://bandit.readthedocs.io/en/latest/plugins/b101_assert_used.html#module-bandit.plugins.asserts

Right now the only violations are in tests. It would be better to use the bandit yaml configuration to disable the check just for tests, but it seems we don't have the ability to use bandits yaml configuration that would allow us to do that. I've filed an issue:

https://github.com/trunk-io/plugins/issues/647

jmontleon: Learned how to use bandit.yaml config on the linked issue!

jwmatthews (approved):

PR #23, opened 2024-02-07
add querying to incident store

savitharaghunathan:
**Assumptions:**
1. Sample Analysis reports have two reports - initial & solved
2. App.yaml is provided for all apps with details of git start and end commit
3. Incident store is loaded once with all the apps at first, it compares initial and solved reports and creates solved issues

**How to run**
1. To load incident store and create necessary files
` ./run_cli.py load samples/analysis_reports` 

PrereQ:
In order to run the following queries, make sure incident store is loaded and available

2. To query solved incidents and get a patch for a given ruleset and violation
`./run_cli.py patch quarkus/springboot javaee-pom-to-quarkus-00010 `

3. To find common violations,
`./run_cli.py common quarkus/springboot javaee-pom-to-quarkus-00010 `

It should generate two patches one from eap-coolstore and another from helloworld-mdb


jwmatthews (approved):
I see we have an increasing number of good things landing in this PR.   I'd prefer for us to move quick and break things opposed to going slower and cautious.  My vote is to merge this once @savitharaghunathan feels good, then we can all build on top of what is in this PR.

PR #32, opened 2024-02-17
JMS to Smallrye example: Zero Shot and Few Shot

jwmatthews:
Note, there is an issue with calculating a diff for the solved example which needs to be solved in a follow up PR


savitharaghunathan (approved):
visiack

fabianvf (approved):

PR #33, opened 2024-02-18
Adding tests for scm.GitDiff and update notebook so git diff now works

jwmatthews:
Create a unit test file for working with scm.GitDiff so we can work with various branches and extract a diff.

The issue I had was I wasn't working with commit sha's correctly so my prior attempts at extracting a diff were failing.  After getting all of this working with the unit test I updated the notebook example so it can pass in the solved example as a diff to the LLM


savitharaghunathan (approved):
/lgtm

PR #37, opened 2024-02-22
adding tests, adding ticketmonster, updated incident store 

savitharaghunathan:
* Updated git tests
* Added ticketmonster to the solved incidents
* updated incident store to match the recent fetch script repo names


jwmatthews:
@savitharaghunathan when you create the example with multiple solved examples consider if we want to account for that in the prompt, I think the multi-example currently is creating a diff of multiple files but not sure we capture that in the prompt

We may want to do something like:

Here are 3 similar solved examples:
Example #1:  Filename 1
Diff 1

Example #2: Filename 2
Diff 2

Example #3: Filename 3
Diff 3

notebooks/jms_to_smallrye_reactive/output/gpt-3.5-turbo-16k/helloworldmdb/custom-ruleset/jms-to-reactive-quarkus-00010/few_shot/05-solved_examples/template.txt
```diff
@@ -0,0 +1,249 @@
...
+
+    # Input information
+    ## Issue found from static code analysis of the Java EE code which needs to be fixed to migrate to Quarkus
+    Issue to fix:  "Enterprise Java Beans (EJBs) are not supported in Quarkus. CDI must be used.
+ Please replace the `@MessageDriven` annotation with a CDI scope annotation like `@ApplicationScoped`."
+
+
+    ## Solved Example Git Diff 
+    This diff of the solved example shows what changes we made in past to address a similar problem. 
+    Please note that this might have many solved examples that may be relevant to the issue found in the Java EE code.
```
jwmatthews, line 33: I see we give some guidance here that this may include multiple solved examples.  In future we may want to iterate to be a little more explicit with example #1, example #2, example #3.   I don't know for sure, just an idea to consider.

savitharaghunathan:
> @savitharaghunathan when you create the example with multiple solved examples consider if we want to account for that in the prompt, I think the multi-example currently is creating a diff of multiple files but not sure we capture that in the prompt
> 
> We may want to do something like:
> 
> Here are 3 similar solved examples: Example #1: Filename 1 Diff 1
> 
> Example #2: Filename 2 Diff 2
> 
> Example #3: Filename 3 Diff 3

@jwmatthews do we want this to be dynamically populated based on the number of examples? Right now, patches is a list that contains only the diff of solved incidents. Should I include filename as well?

jwmatthews:
> > @savitharaghunathan when you create the example with multiple solved examples consider if we want to account for that in the prompt, I think the multi-example currently is creating a diff of multiple files but not sure we capture that in the prompt
> > We may want to do something like:
> > Here are 3 similar solved examples: Example #1: Filename 1 Diff 1
> > Example #2: Filename 2 Diff 2
> > Example #3: Filename 3 Diff 3
> 
> @jwmatthews do we want this to be dynamically populated based on the number of examples? Right now, patches is a list that contains only the diff of solved incidents. Should I include filename as well?

What if we pick an example where we have 3 solved examples, then we explicitly pass in info for example1, example2, example3 the filename and diff.

That is probably a small change and allows us to get a feel for what it'd look like to have 3 solved examples in a prompt.  I think going through to dynamically change the number of examples will be a bit of a PITA and not worth tackling that in the notebook.

My main thought was the LLMs could get confused with us just dumping in several diffs without enough context to distinguish each one being unique.

notebooks/jms_to_smallrye_reactive/output/gpt-3.5-turbo-16k/helloworldmdb/custom-ruleset/jms-to-reactive-quarkus-00010/few_shot/05-solved_examples/template.txt
```diff
@@ -0,0 +1,254 @@
...
+    ## Issue found from static code analysis of the Java EE code which needs to be fixed to migrate to Quarkus
+    Issue to fix:  "Enterprise Java Beans (EJBs) are not supported in Quarkus. CDI must be used.
+ Please replace the `@MessageDriven` annotation with a CDI scope annotation like `@ApplicationScoped`."
+
+
+    ## Solved Example Git Diff 
+    This diff of the solved example shows what changes we made in past to address a similar problem. 
+    Please note that this might have many solved examples that may be relevant to the issue found in the Java EE code. 
+    Please consider this heavily in your response. 
+    Below are the 2 similar solved examples and use them as a guidance to provide solution to the given issue.
```
savitharaghunathan, line 35: @jwmatthews I made the necessary changes. PTAL.

savitharaghunathan: I am specifying the two examples as example 1 and 2 with the diff

jwmatthews (approved):
LGTM

PR #38, opened 2024-02-27
Update JMS Output to run against the 5 common rule violations between both apps

jwmatthews:
Adds a new notebook example: 06_few_shot_with_analysis_jms-to-reactive-ALL.ipynb

This will find the common rule violations between the 2 sample apps and then iterate over all the common rules.
In process of this needed to tweak how we parse the output to account for XML files as well as Java files.


notebooks/jms_to_smallrye_reactive/06_few_shot_with_analysis_jms-to-reactive-ALL.ipynb
```diff
@@ -0,0 +1,754 @@
...
+    "        result,\n",
+    "    )\n",
+    "    print(f\"Completed processing of rule: {rule_name}\")"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Should we consider iterating over fixes one at a time?\n"
```
savitharaghunathan, line 729: I was contemplating about this a while ago. May be worth discussing this with the team.

savitharaghunathan (approved):
/lgtm

PR #40, opened 2024-02-28
Update pydantic notebook to output results/traceback to file

jmontleon:


jmontleon:
@jwmatthews I added the non-pydantic text-gen runs as well. CodeLLama 13b and 34B and Mistral 7b to a lesser extent were probably the more interesting of those.

I think I've been seeing better results in the web ui than over the API so I'm trying to more closely match those settings. top_p=0.9 (instead of 1), temperature=0.01 (instead of 0.1), etc. I think it's improving the results. I'll continue to tweak and look at adjusting the prompt. If I make any concrete break through I'll update the results in a new PR.

jwmatthews (approved):

PR #41, opened 2024-03-06
Add JPA Example

jwmatthews:
This PR is mainly updates to how we automate the analysis of sample app data so we can add more custom rules and then working through 2 notebooks for JPA.   

- Adds 2 notebooks to show JPA examples
- Added a new custom rule for JPA 
- Added a new custom rule for EJB-Security
- Change analysis runs so they will use any custom rules we place in 'samples/custom_rules'
- Splits out the fetch_and_analyze to separate scripts 
- Removes unused/old files like application_hub.py
- Update incident_store.py for sample application paths so unit tests pass
- Update analysis_reports so they account for the new custom rules
- Automate the creation of the app.yaml so as we add new apps, or delete the analysis_reports directory we can safely recreate this file for incident_store


samples/config.py
```diff
@@ -0,0 +1,62 @@
...
+    "ejb-remote": "sample_repos/ejb-remote",
+    "ejb-security": "sample_repos/ejb-security",
+    "tasks-qute": "sample_repos/tasks-qute",
+    "greeter": "sample_repos/greeter",
+}
+
+sample_target_apps = {
+    "eap-coolstore-monolith": "sample_repos/eap-coolstore-monolith",
+    "ticket-monster": "sample_repos/ticket-monster",
+    "kitchensink": "sample_repos/kitchensink/kitchensink",
```
savitharaghunathan, line 54: Suggestion: We can use sample_source_apps instead as they are one and the same. May be rename it to sample_apps

savitharaghunathan (approved):
/lgtm

samples/config.py
jwmatthews: Good point, the need to break these out went away as we bring sample apps into konveyor-ecosystem.

Thank you, will update in follow up PR.

PR #43, opened 2024-03-06
Added Postgres Incident Store & Prompt Builder

JonahSussman:
Please feel free to tear this apart! It still needs tests and a containerized version of Postgres. The schema for REST API (which is just one endpoint really right now) is currently in JSON Schema, but we can switch to Pydantic if that fulfills the requirements as well


kai/data/templates/solved_example_none.txt
```diff
@@ -0,0 +1,4 @@
+## Solved Example
```
savitharaghunathan, line 1: +1 for thinking about no solved example case :)

kai/incident_store_advanced.py
```diff
@@ -0,0 +1,855 @@
...
+
+  def get_max_tokens(self) -> int:
+    return self.max_tokens
+  
+
+  def get_dimension(self) -> int:
+    return self.dimension
+  
+
+class EmbeddingOpenAI(EmbeddingProvider):
```
savitharaghunathan, line 90: if we choose to use different model, would this function require any changes? If so, should we make it generic if possible.

kai/incident_store_advanced.py
```diff
@@ -0,0 +1,855 @@
...
+
+  def insert_and_update_from_report(self, app: Application, report: Report):
+    """
+    Returns: (number_new_incidents, number_unsolved_incidents,
+    number_solved_incidents): tuple[int, int, int]
+    """
+    # FIXME: Only does stuff within the same application. Maybe fixed?
+
+    # create entries if not exists
+    # reference the old-new matrix
```
savitharaghunathan, line 631: Like this matrix. best explained 👏

kai/incident_store_advanced.py
```diff
@@ -0,0 +1,855 @@
...
+
+
+if __name__ == '__main__':
+  psqlis = PSQLIncidentStore(
+    config_filepath='database.ini', 
+    config_section='postgresql',
+    emb_provider=EmbeddingNone()
+  )
+
+  old_cmt_commit = 'c0267672ffab448735100996f5ad8ed814c38847'
```
savitharaghunathan, line 830: Suggestion: move the following to a test file

kai/data/misc/jonah-notes.md
```diff
@@ -0,0 +1,24 @@
+# Jonah Notes
```
jwmatthews, line 1: @JonahSussman do you want to commit this?

I'm not against committing notes/questions/etc to help show our thinking if this is what you want.

Just wanted to check if this was included by accident

jwmatthews:
@JonahSussman would you be able to address the trunk fmt/check concerns?

kai/data/misc/jonah-notes.md
JonahSussman: Yep, was committed in error. Removed

kai/incident_store_advanced.py
JonahSussman: We can make a new EmbeddingProvider if we want to swap out embeddings, it'd (hopefully :crossed_fingers:) be not too difficult.

PR #44, opened 2024-03-06
Adding an example with EJB Remote rule

jwmatthews:
- Added EJB Remote notebook
- Updated analyze_apps.py to use just a 'sample_apps' instead of 'sample_source_apps' and 'sample_target_apps'


savitharaghunathan (approved):
/lgtm

PR #49, opened 2024-03-08
Updated analysis report info to run a full check with dependencies

jwmatthews:
We were trying to limit the analysis time by using -m source-only that will skip dependency analysis.  Then we learned there are a few rules ([example]( https://github.com/konveyor-ecosystem/kai/blob/main/notebooks/ejb_remote/analysis_report/custom_rules/01-jms-to-reactive-quarkus.windup.yaml#L24C1-L32C64))  which will not properly run if we do this.  My current understanding is that -m source-only is probably 'good enough' for our needs, but to be thorough I'm omitting it for now so we can experiment with the full amount of incidents possible.

Rough idea of performance cost when running [analyze_apps.py](https://github.com/konveyor-ecosystem/kai/blob/main/samples/analyze_apps.py):
- with ```-m source-only```:  ~70 minutes on M1Max
- without:  ~250 minutes on M1Max


savitharaghunathan (approved):
/lgtm

PR #51, opened 2024-03-08
Add Makefile with targets for postgres and server

fabianvf:
also various cleanups required to run this locally, fixing of relative or machine-specific directories, etc


samples/analyze_apps.py
```diff
@@ -67,7 +67,7 @@ def analyze(source_dir, name, target):
     full_output_dir = os.path.join(os.getcwd(), f"analysis_reports/{name}/{target}")
     ensure_output_dir_exists(full_output_dir)
     print(f"Analyzing '{source_dir}', will write output to '{full_output_dir}'")
-    cmd = f'time ./bin/kantra analyze -i {source_dir} -m source-only -t "quarkus" -t "jakarta-ee" -t "jakarta-ee8+" -t "jakarta-ee9+" -t "cloud-readiness" --rules ./custom_rules -o {full_output_dir} --overwrite'
+    cmd = f'time ./bin/kantra analyze -i {source_dir} -m source-only -t "quarkus" -t "jakarta-ee" -t "jakarta-ee8+" -t "jakarta-ee9+" -t "cloud-readiness" --rules {os.path.dirname(__file__)}/custom_rules -o {full_output_dir} --overwrite'
```
jwmatthews, line 70: @fabianvf let's update this to remove '-m source-only'

I submitted the below recently to go back to removing source-only:

https://github.com/konveyor-ecosystem/kai/pull/49/files#diff-fb1c6627c87cd8284545997fcc69c4b5d56d1b5be5c6be4038111c56b899fc1b

I submitted that as @shawn-hurley pointed out that as of right now [one of our custom rules](https://github.com/konveyor-ecosystem/kai/blob/main/notebooks/ejb_remote/analysis_report/custom_rules/01-jms-to-reactive-quarkus.windup.yaml#L24C1-L32C64) wouldn't be caught with source-only, I'm unsure what other Quarkus rules may be impacted as well, so removed the source-only until we know more.

jwmatthews (changes requested):
lgtm, just need to account for removing '-m source-only' in analyze_apps.py

samples/analyze_apps.py
fabianvf: done

jwmatthews (approved):

PR #56, opened 2024-03-11
Seed db

savitharaghunathan:
Run `make load-data`

### DIFF
--- a/kai/incident_store_advanced.py
+++ b/kai/incident_store_advanced.py
@@ -1,3 +1,4 @@
+import argparse
 import datetime
 import json
 import os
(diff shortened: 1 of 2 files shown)


kai/incident_store_advanced.py
```diff
@@ -925,51 +972,143 @@ def accept_solution_from_id(
...
+        return app_data
+
+
+if __name__ == "__main__":
+
+    psqlis = PSQLIncidentStore(
+        config_filepath="database.ini",
+        config_section="postgresql",
+        emb_provider=EmbeddingNone(),
+        drop_tables=True,
```
JonahSussman, line 1111: Probably don't want to drop the tables each time

JonahSussman (changes requested):

kai/incident_store_advanced.py
savitharaghunathan: Makes sense. I added make target for loading data and we can change it via args

JonahSussman:
Looks good to me!

JonahSussman (approved):

PR #64, opened 2024-03-12
Ensure we are on the initial branch prior to running analysis

### LINKED ISSUE #63: No longer any difference between `initial` and `solved` analysis reports
As far as I can see, there is no longer any difference in the analysis reports between the `initial` and `solved` versions, except for `ticket-monster`. I think this regression was caused by https://github.com/konveyor-ecosystem/kai/commit/300dd5bccc7461f5f2e707265ca3ec43711812a4.

I ran the following script in `samples/analysis_reports`:

```bash
#!/bin/bash

for dir in */; do
    if [[ -f "$dir/initial/output.yaml" && -f "$dir/solved/output.yaml" ]]; then
        echo "Diff for directory: $dir"
        git diff --no-index "$dir/initial/output.yaml" "$dir/solved/output.yaml"
    else
        echo "Files not found in directory: $dir"
    fi

    read -p "Press any key to continue to the next directory..." -n 1 -r
    echo
done
```

And got the following output:

[output.txt](https://github.com/konveyor-ecosystem/kai/files/14574875/output.txt)

While it appears that a lot changed in `ticket-monster`, really only one incident was solved.

jwmatthews:
Fixes #63

### DIFF
--- a/samples/analyze_apps.py
+++ b/samples/analyze_apps.py
@@ -8,11 +8,20 @@ from config import repos, sample_apps
 
 
 def analyze_apps():
+    # Ensure we are on the initial branch for each repo
+    for repo in repos:
+        print(f"Switching to {repos[repo][1]} branch for {repo}...")
+        os.chdir(f"sample_repos/{repo}")
+        if repos[repo][1] is not None:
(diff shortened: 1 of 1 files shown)


jwmatthews:
Closing in favor of PR #66

PR #66, opened 2024-03-12
Updated analysis output to fix issue of initial & solved being same

jwmatthews:
Fixes issue #63

This will ensure we check out the main branch prior to running the initial analysis.
Issue was we had quarkus checked out sometimes on prior runs, so initial/solved ended up both running on quarkus branch


jwmatthews:
Not yet ready, @savitharaghunathan noticed some oddities with eap-coolstore-monolith and cmt, their corresponding output.yamls are not in the expected location.  Looks like a problem with the script.

jwmatthews:
Close in favor of #67

PR #67, opened 2024-03-13
Ensure we are on appropriate branch for initial or solved analysis

jwmatthews:
Fixes issue #63 

The core of the problem is that after we run [fetch_apps.py](https://github.com/konveyor-ecosystem/kai/blob/main/samples/fetch_apps.py) we are on the 'solved' branch.

[analyze_apps.py](https://github.com/konveyor-ecosystem/kai/blob/main/samples/analyze_apps.py) was not adjusting the checked out branch to match the 'initial' branch, so we were running analysis twice on the 'solved' branch but storing the output as 'initial' and as 'solved'.

The fix added is to checkout out the 'initial' branch prior to first round of analysis


savitharaghunathan (approved):
VisualACK

PR #74, opened 2024-03-14
Unit tests for incidentstore

savitharaghunathan:


tests/test_data/data/database.ini
JonahSussman, line 1: The incident store has a config field in its constructor of you want to use that instead of creating a new file

JonahSussman (approved):

kai/incident_store_advanced.py
```diff
@@ -973,9 +974,9 @@ def accept_solution_from_id(
     def accept_solution_from_diff():
         pass
 
-    def load_store(self):
+    def load_store(self, path: str):
         # fetch the output.yaml files from the analysis_reports/initial directory
-        path = "samples/analysis_reports"
+        # path = "samples/analysis_reports"
```
fabianvf, line 979: can just delete this line

fabianvf (commented):
LGTM, minor non-blocking nit

tests/test_data/data/database.ini
savitharaghunathan: I thought of that but it doesn't provide a separation between test data and the real one. Its a copy of what we have but with an option to modify the config just for testing. I can switch back if you think its redundant.

kai/incident_store_advanced.py
savitharaghunathan: deleted the line.

savitharaghunathan:
solved https://github.com/konveyor-ecosystem/kai/issues/70

tests/test_data/data/database.ini
JonahSussman: All good!

PR #77, opened 2024-03-15
Set aiohttp version based on python version

jmontleon:
Fedora 39 uses python 3.12. aiohttp 3.8.6 does not work with it.


fabianvf (approved):
LGTM

PR #79, opened 2024-03-19
Updates to run on MacOS arm64

### LINKED ISSUE #76: Cannot install dependencies on M series Mac
Several installed libraries are exclusive to x86_64, Linux, or both, [for example](https://github.com/openai/triton?tab=readme-ov-file#compatibility). Luckily these are dependencies of dependencies, so we could switch to a package management system like Poetry, [which supports conditionally installing dependencies based on the platform](https://python-poetry.org/docs/dependency-specification/#python-restricted-dependencies). While I don't foresee people running the server on a Mac, it might be good for getting more developers on board.

Specifically, it appears that `torch==2.2.0` is not well supported for M series Macs. One solution, for now, is just to install `torch==2.0.0`, ignore `torchvision==0.17.0` and `triton==2.0.0`, and let `pip` figure out the right dependencies. However, `pip freeze` will always output every installed package making it difficult to maintain. 

Another reason to pivot to a more advanced package management system is that, unless the user wants to use embeddings, there's no reason to install all the machine learning stuff on a person's machine.

jwmatthews:
Fixes #76


jmontleon:
@jwmatthews curious if this would work for you:
https://gist.github.com/jmontleon/6f9334263f44354ac28e22973347286a

jwmatthews:
@jmontleon I will try that, thank you

jwmatthews:
Thank you @jmontleon the suggestions worked great!

I ended up trying them in a different branch, so just going to close this PR and use PR#80

PR #80, opened 2024-03-19
Update to work with MacOS

### LINKED ISSUE #76: Cannot install dependencies on M series Mac

jwmatthews:
Fixes #76

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,10 +1,13 @@
 CONTAINER_RUNTIME = podman
 
+CWD := $(shell pwd)
+KAI_PYTHON_PATH="$(CWD)/kai:$(PYTHONPATH)"
+
 run-postgres:
 	$(CONTAINER_RUNTIME) run -it -v data:/var/lib/postgresql/data -e POSTGRES_USER=kai -e POSTGRES_PASSWORD=dog8code -e POSTGRES_DB=kai -p 5432:5432 docker.io/pgvector/pgvector:pg15
 
 run-server:
-	python ./kai/server.py
+	PYTHONPATH=$(KAI_PYTHON_PATH) python ./kai/server.py
 
 load-data:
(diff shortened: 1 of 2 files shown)


jwmatthews:
@JonahSussman @savitharaghunathan please see if this works for you in a clean venv.

@jmontleon showed how we can include or ignore specific requirements for 'darwin'.

I tried this on MacOS arm64 with python 3.11 and python 3.12 and looking good.

JonahSussman:
Don't have an arm mac to test but lgtm

jmontleon (approved):
ACK

jwmatthews:
@JonahSussman sorry, I meant please try these requirements.txt out on your Linux box.

In theory this same requirements.txt should work on Linux or Mac.

JonahSussman:
Should have used the logic part of my brain! Yeah, there are no issues using it on my x86_64 machine

JonahSussman (approved):

savitharaghunathan (approved):
Tested the changes on my macOS/intel and it works.

PR #84, opened 2024-03-20
Output for coolstore analysis

jwmatthews:
This is the output from running PR #83 

Experimenting with separating the PRs to make it easier to review the scripts/code and not get lost in noise of the analysis output.

2 kinds of output in here:

- Kantra output
- Markdown output from https://github.com/konveyor-ecosystem/kai/blob/main/kai/report.py#L96

I created the Markdown output to learn/play a little with the yaml output and to make it easier to share info with folks in GitHub of the specific analysis incidents we found in an app.


savitharaghunathan (approved):
visual ack

PR #86, opened 2024-03-20
Add preliminary batching support

JonahSussman:
We went with a naive approach of taking the list of incidents, and looping over them using the existing logic to process each incident, but passing the updated file forward.

### DIFF
--- a/kai-service/mock-client.py
+++ b/kai-service/mock-client.py
@@ -45,76 +45,171 @@ async def main():
         #     "analysis_message": "main function is too complicated!",
         # }
 
-        x = {
-            "application_name": "test_app",
-            "violation_name": "jms-to-reactive-quarkus-00010",
-            "ruleset_name": "kai/quarkus",
-            "incident_snip": """  1  /*
-  2   * JBoss, Home of Professional Open Source
-  3   * Copyright 2015, Red Hat, Inc. and/or its affiliates, and individual
-  4   * contributors by the @authors tag. See the copyright.txt in the
-  5   * distribution for a full listing of individual contributors.
-  6   *
-  7   * Licensed under the Apache License, Version 2.0 (the \"License\");
-  8   * you may not use this file except in compliance with the License.
-  9   * You may obtain a copy of the License at
- 10   * http://www.apache.org/licenses/LICENSE-2.0
- 11   * Unless required by applicable law or agreed to in writing, software
- 12   * distributed under the License is distributed on an \"AS IS\" BASIS,
- 13   * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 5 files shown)


JonahSussman:
Notes for later:

- If we want to add diff support in the future, we will need to more intelligently handle line numbers. 
- I'm starting to hate myself for the way I did the prompt builder (hard to have an arbitrary number of arguments). 
- May have to look again at how we handle marking incidents as solved in the store.

kai-service/mock-client.py
```diff
@@ -176,35 +271,110 @@ async def main():
...
   }""",
-            "line_number": 36,
-            "analysis_message": """Enterprise Java Beans (EJBs) are not supported in Quarkus. CDI must be used.
-Please replace the `@MessageDriven` annotation with a CDI scope annotation like `@ApplicationScoped`.""",
+            "application_name": "test_app",
+            "incidents": [
+                {
+                    "violation_name": "jms-to-reactive-quarkus-00010",
+                    "ruleset_name": "kai/quarkus",
+                    "incident_snip": """  1  /*
```
jwmatthews, line 279: Just a FYI, I think we should drop incident_snip.

I don't believe we want to send incident_snip in to our API or consider it with the prompt builder.

fabianvf: could we fix this in a followup for all the endpoints?

kai/server.py
```diff
@@ -331,6 +350,49 @@ async def ws_get_incident_solution(request: Request):
...
+    for incident in request_json["incidents"]:
+        incident["file_name"] = request_json["file_name"]
+        incident["file_contents"] = current_file_contents
+        incident["application_name"] = request_json["application_name"]
+
+        llm_output = get_incident_solution(incident, False)
+        content = parse_file_solution_content(llm_output.content)
+
+        total_reasoning.append(content.reasoning)
+        current_file_contents = content.updated_file
```
jwmatthews, line 386: This makes sense for a first cut

jwmatthews (approved):
I'm not setup to test this out yet, still working up a client.

Reading over the changes, this looks good and makes sense.

kai-service/mock-client.py
jwmatthews: Oh yeah, dont need to address now, more just information. 
@JonahSussman and I spoke about codeSnip yesterday and there was some confusion if we should use it.

My thinking is we should ignore codeSnip for Kai purposes and assume the input file contents we are sent are what we want to focus on.

All good to merge this now as-is and we can adjust for codeSnip later.

fabianvf:
@jwmatthews did you want to test it before merge or should I just go ahead?

jwmatthews:
Go ahead and merge now

I will be in position to test this later today/tomorrow morning, if I find issues we can update as we go.

PR #92, opened 2024-03-21
Restore run_tests.sh to run without an errors

jwmatthews:
I wasn't able to run unit tests today, noticed 2 failures that needed a tweak to address

### DIFF
--- a/tests/test_incident_store.py
+++ b/tests/test_incident_store.py
@@ -243,7 +243,7 @@ class TestIncidentStore(unittest.TestCase):
             "quarkus/springboot", "javaee-pom-to-quarkus-00010"
         )
         self.assertIsNotNone(patches)
-        self.assertEquals(len(patches), 1)
+        self.assertEqual(len(patches), 1)
         self.incident_store.cleanup()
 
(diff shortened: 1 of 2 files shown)


tests/test_incident_store.py
```diff
@@ -243,7 +243,7 @@ def test_find_solved_issues(self):
             "quarkus/springboot", "javaee-pom-to-quarkus-00010"
         )
         self.assertIsNotNone(patches)
-        self.assertEquals(len(patches), 1)
+        self.assertEqual(len(patches), 1)
```
savitharaghunathan, line 246: @jwmatthews Should we remove this out of tests in favor of https://github.com/konveyor-ecosystem/kai/blob/main/tests/test_incident_store_advanced.py

we are not using `incident_store.py` anymore to load data.

jwmatthews: What if we merge this just to keep run_tests.sh working.  
Then we follow up with a PR to remove both incident_store.py and this test_incident_story.py?

savitharaghunathan (approved):
/lgtm

PR #125, opened 2024-03-28
Added support for additional model args

JonahSussman:


kai/config.toml
```diff
@@ -21,8 +21,13 @@ password = "dog8code"
 #   OpenAI GPT 4
 #     provider = "OpenAI"
 #     args = { model_id = "gpt-3.5-turbo" }
-provider = "IBMGranite"
-args = { model_id = "ibm/granite-13b-chat-v2" }
+# provider = "IBMGranite"
+# args = { model_id = "ibm/granite-13b-chat-v2" }
```
jwmatthews, line 25: Is the intent to not have a model configured by default?

JonahSussman: Mistype, fixing

jwmatthews (approved):
Should we also add in `max_new_tokens` and `min_new_tokens`?

JonahSussman:
Sounds good, will add

PR #128, opened 2024-03-29
Process non-java files in demo and log what model is actually used in a result

jwmatthews:
With this PR we will run the demo against pom.xml, .properties, and other non-java files.

This will also write a line in each 'result' file to confirm the model used.
In this example I was configured with:

```
INFO - 2024-03-29 09:42:17,795 - [server.py:571 -                  app() ] - Config loaded: {'embeddings': {'todo': True},
 'models': {'args': {'model_id': 'codellama/codellama-34b-instruct'},
            'provider': 'IBMOpenSource'},
```

Yet we can see it defaulted to `meta-llama/llama-2-13b-chat`.   PR #127 will help to improve this so in future we won't silently default to a different model than what is configured

```
$ head pom.xml.llm_result 
Model ID: meta-llama/llama-2-13b-chat


# Java EE to Quarkus Migration

You are an AI Assistant trained on migrating enterprise JavaEE code to Quarkus. I will give you an example of a JavaEE file and you will give me the Quarkus equivalent.
```

### DIFF
--- a/example/run_demo.py
+++ b/example/run_demo.py
@@ -171,7 +171,9 @@ def write_to_disk(file_path, updated_file_contents):
         llm_result_path = f"{intended_file_path}.llm_result"
         KAI_LOG.info(f"Writing llm_result to {llm_result_path}")
         try:
+            model_id = updated_file_contents.get("model_id", "unknown")
             with open(llm_result_path, "w") as f:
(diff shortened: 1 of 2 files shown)


jmontleon (approved):
ACK

PR #141, opened 2024-04-03
Update IBM / OpenSource default values

jmontleon:
Juanma stated that the results improved after making these changes in combination with prompt changes.
https://github.com/konveyor-ecosystem/coolstore/pull/19

temperature 0.05 is the minimum IBM accepts and will probably generally result in more reproducible results, but open to discussing adjusting any of these.


fabianvf (approved):
LGTM

Just wondering, in the long-term would we want individual models to be tuned differently, or do we expect these parameters to perform better across the board? if the former, maybe we want preconfigured arg blocks for models we want to support in the config.toml

PR #142, opened 2024-04-03
Update default preamble

jmontleon:
Updating the prompt along with values for the model appears to have resulted in improved results.
https://github.com/konveyor-ecosystem/coolstore/pull/19

Creating a PR in case others want to try to reproduce more positive results.

Open to hearing others results along with ideas on further improving this.


JonahSussman:
LGTM

PR #144, opened 2024-04-03
Add demo mode

fabianvf:
- Adds demo mode (invoked like `DEMO_MODE=true make run-server`) which will cache request/responses (or use the cached request/response if already present)
- Added additional debug logging

If a response has already been recorded, and another comes in for the same file but with different request contents (ie, the source file changed or something), it will error. If the difference is intentional and you want to rewrite the cassette, just delete the corresponding file in `kai/data/vcr/<application>/<model_id>/` and it will re-record the cassette on the next run.

I recommend reviewing the first commit. The second commit contains all the committed cassettes and makes for pretty poor reading material.


kai/server.py
```diff
@@ -51,6 +54,37 @@
...
+            ],
+            record_on_exception=False,
+        )
+        with my_vcr.use_cassette(f"{filename}.yaml", filter_headers=["authorization"]):
+            yield
+    else:
+        yield
+
 
 def load_config():
```
JonahSussman, line 89: `load_config`, `load_templates`, `load_template`, and  `generate_prompt` are unused, I think. It should be safe to delete

fabianvf: :+1:. will leave that for a separate PR though

jwmatthews (approved):
I'm running this now with IDE talking to backend, looking good.

I see new requests go through as normal, then a subsequent call returns quickly as it's cached.

jwmatthews:
I am seeing issues running with this and the IDE.  I don't know enough yet where the problems are coming from.
I'm taking notes here:  https://github.com/konveyor-ecosystem/kai/issues/147

PR #159, opened 2024-04-08
Record all responses

fabianvf:
Sets VCR to record responses in or out of demo mode, but only play them back in demo mode. This is to make re-recording easier as changes are made, plus should aid with debugging. 

Updates the output instructions to allow for the case when the file should be removed to make that consistently parsable.

Makes the matching a little less strict, if content-length/body don't match it will still send the original request. This will mean changing the prompt no longer changes the response, so re-recording will be necessary.


jmontleon (approved):
ACK

PR #166, opened 2024-04-11
Adjust rule text to improve mistral results

jmontleon:
I don't know if we want to be manually changing this file, but with this change I was able to get `mixtral-8x7b-instruct-v0-1` to add  the import to `OrderService.java`.

If this is an acceptable approach for now, I believe there are other places in the output.yaml we could probably update this text.

### DIFF
--- a/example/analysis/coolstore/output.yaml
+++ b/example/analysis/coolstore/output.yaml
@@ -3004,7 +3004,7 @@
           name: Stateless
           package: com.redhat.coolstore.service
       - uri: file:///tmp/source-code/src/main/java/com/redhat/coolstore/service/OrderService.java
-        message: "Stateless EJBs can be converted to a cdi bean by replacing the `@Stateless` annotation with a scope eg `@ApplicationScoped`\n "
+        message: "Stateless EJBs can be converted to a cdi bean by importing `jakarta.enterprise.context.ApplicationScoped` and replacing the `@Stateless` annotation with a scope eg `@ApplicationScoped`\n "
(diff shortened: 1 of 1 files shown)


fabianvf:
@jmontleon we should probably check in the updated cached request that you saw better performance with too

jwmatthews (changes requested):
This is great for testing purposes to learn what will help, but I worry direct changes to the output is wrong approach that will hurt us (likely to miss these changes as we update analysis in future). 

I'd ask we do this only to test, then once we know what changes are desired we address those changes in the rules so we can refresh the analysis.

For example for this rule we would update:
- https://github.com/konveyor/rulesets/blob/main/default/generated/quarkus/201-cdi-to-quarkus.windup.yaml#L93

As to updating the builtin rules that ship with Kantra and testing.... we should ask @jmle and @pranavgaikwad for their suggestion of best way we can overwrite what we use and then we can update our analysis scripts to use this method.

Be aware there appears to be an open PR that is updating the rules so we may want to coordinate with @pranavgaikwad to see https://github.com/konveyor/rulesets/pull/71/files#diff-b4e32d57989921cb7835f62eddc866e184ba37ed787e29af0c90803f3938c09bR93 merge and then we make our change on top.

jwmatthews:
This can help us to override the Kantra default rules and allow us to have a copy in this repo so we can tweak:
https://github.com/konveyor-ecosystem/kai/pull/167/files

jwmatthews:
@jmontleon we have the output you found that helps now in:
https://github.com/konveyor-ecosystem/kai/blob/main/example/analysis/coolstore/output.yaml#L2628-L2637

We can tweak this message for subsequent analysis runs by editing the rule file:
https://github.com/konveyor-ecosystem/kai/blob/main/example/default_rules/quarkus/201-cdi-to-quarkus.windup.yaml#L99

and then re-running:
https://github.com/konveyor-ecosystem/kai/blob/main/example/analyze.sh

jmontleon:
Thank you @jwmatthews !

PR #167, opened 2024-04-11
For demo analysis of coolstore, override the default rules with our own version we can tweak

jwmatthews:
This helps us to ignore the default rules in Kantra.
Then we can tweak what we want for testing and regenerate the analysis.

Assumption is as we learn what tweaks we want we will contribute those back to the main rulesets directory.


fabianvf (approved):
Nice!

savitharaghunathan (approved):
Thank you :)

PR #171, opened 2024-04-18
Naive incident store fix

JonahSussman:
NOTE: With the new solved incidents being included, the prompts can get too long and might fail. I added a piece of code to remove the solved incidents when the request fails. A better fix would be to create a new method on the model provider base class to parse _why_ it failed. Feel free to block this merge if you guys want to go that approach.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,6 +4,7 @@ CWD := $(shell pwd)
 KAI_PYTHON_PATH="$(CWD)/kai:$(PYTHONPATH)"
 LOGLEVEL ?= info
 NUM_WORKERS ?= 8
+DROP_TABLES ?= False
 
 run-postgres:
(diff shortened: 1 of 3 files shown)


jwmatthews (changes requested):
Out of caution, I'd like to ask we hold off a few days on merging this into main.  

We are recording the demo now, just incase this alters behavior I'd like to limit churn for any demo blocker bugs we need to address.

In meantime, it'd be nice to run through tests with this and coolstore and get a feel how often requests are failing and see if we are exceeding context windows and decide if we can bring this in with the multiple incidents per request, or do we need to think of something else.

fabianvf:
Waiting for merge makes sense :+1: 

I didn't see an increase in error rates with mixtral using this branch. Results here: https://github.com/konveyor-ecosystem/coolstore/pull/25/files

JonahSussman:
Closing for now. Incident store refactor might make this redundant.

PR #173, opened 2024-04-19
Enable testing with llama3

jwmatthews:
Example of coolstore run here:
https://github.com/konveyor-ecosystem/coolstore/pull/26/files#


savitharaghunathan (approved):
/lgtm

PR #176, opened 2024-04-23
Add Ollama Support

jmontleon:


jwmatthews (approved):
Success using 'mistral', I tried 'llama3' as well but lacked sufficient RAM to run that on my laptop, got an out of memory error after ~70+ minutes.

@jmontleon please also update the kai/config.toml to call out an example of configuring with ollama so folks have a little guidance

jwmatthews (approved):

kai/config.toml
```diff
@@ -20,6 +20,9 @@ password = "dog8code"
 # .   See:  https://github.com/konveyor-ecosystem/kai/issues/172
 #     provider = "IBMOpenSource"
 #     args = { model_id = "meta-llama/llama-3-70b-instruct", max_new_tokens = 2048 }
+#   Ollama
+#     provider = "Ollama"
+#     args = { model_id = "mistral" }
```
jwmatthews, line 25: thank you

jmontleon: You're welcome!

PR #181, opened 2024-05-03
Refactor and create evaluation metrics

JonahSussman:


kai/config.yaml
fabianvf, line 1: just curious, why swap the format?

kai/config.yaml
```diff
@@ -0,0 +1,70 @@
...
+  provider: postgresql
+  args:
+    host: 127.0.0.1
+    database: kai
+    user: kai
+    password: dog8code
+
+models:
+  # **IBM served granite**
+  # ```yaml
```
fabianvf, line 14: wouldn't want the \`\`\`yaml blocks since it's already a yaml file

kai/evaluation.py
```diff
@@ -0,0 +1,145 @@
...
+    for name, config in configs.items():
+        overall_results[name] = []
+
+        ModelProviderClass = ModelProvider.model_from_string(
+            config["models"]["provider"]
+        )
+        ModelProviderClass(**config["models"]["args"])
+
+        # TODO: Make this use directories as opposed to constant
+        for example in BENCHMARK_EXAMPLES:
```
fabianvf, line 86: we'd be grabbing these from `load_benchmarks`

kai/evaluation.py
```diff
@@ -0,0 +1,145 @@
...
+#         Violation:
+#             Incidents:
+#                 uri:
+#                 snip:
+#                 line:
+#                 variables:
+
+
+def load_benchmarks(
+    examples_path=os.path.join(os.path.dirname(__file__), "data", "benchmarks")
```
fabianvf, line 55: Github already called it out, but we shouldn't be doing these function calls in the function arguments

kai/incident_store.py
```diff
@@ -1,748 +1,1091 @@
...
+        incident_snip: Optional[str] = None,
+    ) -> list[Solution]:
         """
-        Load the incident store with the given applications
+        Returns a list of solutions for the given incident. Exact matches only.
         """
+        pass
+
+
+# @dataclass
```
fabianvf, line 232: can probably drop all this

kai/llm_io_handler.py
```diff
@@ -0,0 +1,262 @@
...
+    ] = "single_group",
+    include_solved_incidents: Optional[bool] = True,
+    include_llm_results: bool = False,
+    demo_mode: bool = False,
+):
+    src_file_language = guess_language(file_contents, filename=file_name)
+
+    KAI_LOG.debug(f"{file_name} classified as filetype {src_file_language}")
+
+    # NOTE: Looks worse than it is, `trunk check` mangled the heck out of this
```
fabianvf, line 95: we actually did get the lambdas in so we can drop this comment

fabianvf (changes requested):

kai/config.yaml
JonahSussman: Whoops! I didn't mean to commit that. I was doing some experiments with the config. Some of the model's configs have nested dictionaries, which is a pain to do in TOML. (I only used TOML in the first place because it looks close enough to .ini 😅)

Example:

```toml
[models]
provider = "ChatIBMGenAI"

[models.args]
model_id = "meta-llama/llama-3-70b-instruct"

[models.args.parameters]
max_new_tokens = 2_048
```

```yaml
models:
  provider: ChatIBMGenAI
  args:
    model_id: meta-llama/llama-3-70b-instruct
    parameters:
      max_new_tokens: 2048
```

I'm down to revert it

fabianvf: I think if you're having difficulty with it, it makes sense to change up the format. I can only imagine the options will get more complicated as time goes on. @jwmatthews do you have any opinions?

kai/incident_store.py
```diff
@@ -229,57 +234,195 @@ def find_solutions(
...
-# @dataclass
-# class Incident:
-#     incident_uri: str
-#     incident_snip: str
-#     incident_line: int
-#     incident_variables: dict
-#     application: Application
+# These classes are just for the in-memory data store. Once we figure out the
+# best way to model the incidents, violations, etc... we can remove these
+@dataclass(frozen=True)
```
fabianvf, line 239: I'd create an incident_store directory and move the various implementations to their own files. Just makes it a little easier to read

kai/incident_store.py
```diff
@@ -229,57 +234,195 @@ def find_solutions(
...
+                unsolved_incidents = report_incidents.intersection(store_incidents)
+                number_unsolved_incidents += len(unsolved_incidents)
+
+                solved_incidents = store_incidents.difference(report_incidents)
+                number_solved_incidents += len(solved_incidents)
+                for incident in solved_incidents:
+                    file_path = os.path.join(
+                        repo_path,
+                        unquote(urlparse(incident.uri).path).removeprefix(
+                            "/tmp/source-code/"
```
fabianvf, line 353: what is this directory? If it's just a placeholder I'd name it something like `/placeholder` to make it clear

kai/incident_store.py
```diff
@@ -229,57 +234,195 @@ def find_solutions(
...
+            ruleset = application.rulesets.get(ruleset_name, None)
+            if ruleset is None:
+                continue
+
+            violation = ruleset.violations.get(violation_name, None)
+            if violation is None:
+                continue
+
+            for solved_incident in violation.solved_incidents:
+                if incident_snip != None and solved_incident.snip != incident_snip:
```
fabianvf, line 406: `is not None`

kai/data/benchmarks/configs/default.yaml
```diff
@@ -0,0 +1,6 @@
+model_config:
```
fabianvf, line 1: I think we can get rid of this and just take normal config.tomls

kai/evaluation.py
```diff
@@ -150,23 +242,28 @@ def levenshtein_distance(s1, s2) -> float:
...
-            print(f"Failed to load {filepath}: {e}")
+#         except Exception as e:
+#             print(f"Failed to load {filepath}: {e}")
 
-    print_nicely_formatted_comparison(evaluate(configs))
+#     print_nicely_formatted_comparison(evaluate(configs))
 
 
 if __name__ == "__main__":
-    compare_from_cli()
```
fabianvf: The idea with this was that we could provide different config.tomls and have them run against the benchmarks, to make it super easy to compare two models or a new config without needing to get into the weeds

kai/incident_store.py
JonahSussman: When retrieving uris from the report, some of them had "/tmp/source-code/" as their root path. Unsure where it originates from.

fabianvf: ah I see, weird.

kai/evaluation.py
JonahSussman: Reimplemented!

kai/incident_store.py
JonahSussman: Done!

kai/config.yaml
JonahSussman: Sticking with TOML for now, but I added the capability to load .yaml. It's only a couple lines of code

fabianvf (changes requested):
I think we actually want to use the incident store configurations when running the evaluation, so we can test the impact of solved incidents etc etc. If we are doing that my mistake

kai/server.py
```diff
@@ -433,233 +191,107 @@ async def ws_get_incident_solution(request: Request):
...
-                )
-                KAI_LOG.debug(traceback.format_exc())
-                time.sleep(LLM_RETRY_DELAY)
-        else:
-            KAI_LOG.error(f"{request_json['file_name']} failed to migrate")
-            raise web.HTTPInternalServerError(
-                reason="Migration failed",
-                text=f"The LLM did not generate a valid response: {llm_result}",
-            )
+    # elif os.path.exists(os.path.join(os.path.dirname(__file__), "config.yaml")):
```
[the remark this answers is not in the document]
JonahSussman: Yep, can remove if you want. Added a note in most recent commit

Makefile
```diff
@@ -4,13 +4,14 @@ CWD := $(shell pwd)
...
 	$(CONTAINER_RUNTIME) run -it -v data:/var/lib/postgresql/data -e POSTGRES_USER=kai -e POSTGRES_PASSWORD=dog8code -e POSTGRES_DB=kai -p 5432:5432 docker.io/pgvector/pgvector:pg15
 
 # Can adjust logging level by defining env var: KAI_LOG_LEVEL
 run-server:
-	PYTHONPATH=$(KAI_PYTHON_PATH) gunicorn --timeout 3600 -w $(NUM_WORKERS) --bind localhost:8080 --worker-class aiohttp.GunicornWebWorker 'kai.server:app("$(LOGLEVEL)")'
+	PYTHONPATH=$(KAI_PYTHON_PATH) gunicorn --timeout 3600 -w $(NUM_WORKERS) --bind localhost:8080 --worker-class aiohttp.GunicornWebWorker 'kai.server:app("$(LOGLEVEL)", $(DEMO_MODE))'
 
 load-data:
-	PYTHONPATH=$(KAI_PYTHON_PATH) python ./kai/incident_store_advanced.py  --config_filepath ./kai/config.toml
+	PYTHONPATH=$(KAI_PYTHON_PATH) python ./kai/incident_store.py  --config_filepath ./kai/config.toml
```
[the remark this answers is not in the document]
JonahSussman: Changed in most recent commit

kai/evaluation.py
```diff
@@ -0,0 +1,286 @@
...
+
+
+@dataclass
+class BenchmarkResult:
+    prompt: str
+    llm_result: str
+    similarity: Any
+
+
+def load_single_benchmark_example(full_example_path: str) -> BenchmarkExample:
```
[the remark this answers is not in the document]
JonahSussman: I wasn't, added it to the latest commit

kai/server.py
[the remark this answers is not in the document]
fabianvf: I would go ahead and keep it in personally, since you have the logic to load it from yaml anyway. Wouldn't break anything existing AFAIK?

kai/evaluation.py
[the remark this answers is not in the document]
JonahSussman: Gotcha, I'll file an issue so we can do it later

kai/evaluation.py
```diff
@@ -0,0 +1,286 @@
...
+
+            with open(full_file_path, "r") as f:
+                yaml_incidents = yaml.safe_load(f)
+
+            for yaml_incident in yaml_incidents:
+                incidents.append(Incident.model_validate(yaml_incident))
+
+        else:
+            raise ValueError(
+                f"File must be either `original`, `expected`, or `incidents` in {full_example_path}. Got `{file_name}`."
```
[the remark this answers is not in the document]
JonahSussman: I think it's beneficial to have it. I removed it, updated the structure of the example, and couldn't figure out why it wasn't working. The errors saved me a good chunk of head-desking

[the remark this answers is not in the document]
fabianvf: If we logged it instead would that have been enough to see?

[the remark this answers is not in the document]
JonahSussman: Yeah I'm down, put it in the new commit

Issue #164, opened 2024-04-11
Update configuration so a user may add in arbitrary coordinates for a LLM of their choice.


jwmatthews:
As of early April 2024 we are limiting the choices a user may choose for a model to be based on models available from IBM's BAM service or OpenAI.

We want to provide a means that users may configure arbitrary coordinates for a LLM service so they can test with any LLM of their choice.  

I assume this means we need to open up the Model Provider a little so we can overwrite endpoints at a minimum

jwmatthews:
Related to #148

shawn-hurley:
I believe that this is fixed by the pr merged and with the available config. I am going to close this. If you disagree, please re-open it.

Issue #129, opened 2024-03-29
Change the way we handle model configuration


jmontleon:
Right now we have a set of known models from known providers. This does not seem like it will scale very well though.

Models will probably come and go at a different cadence than our releases and from one provider to another.

We could do something like a have a list of suggested/known/tested/whatever models, but in the end I think it probably make sense to allow any provider and model combination. We can probably document links for API docs for various providers so they can explore the full range of options as well.

jwmatthews:
In regard to IBM BAM models we have been using this page to learn what models are possible:
https://bam.res.ibm.com/docs/models

shawn-hurley:
I think that this completed, if you disagree please re-open.

PR #183, opened 2024-05-09
Update example config/model_provider for "mistralai/mixtral-8x7b-instruct-v01" and add cached data

jwmatthews:
- Update config.toml so the example references the current mixtral version: "mistralai/mixtral-8x7b-instruct-v01"
- Update model_provider to remove the older mixtral models no longer being served from BAM
- Update cached data from run_demo and mock_client for latest mixtral


kai/model_provider.py
```diff
@@ -78,6 +78,7 @@ def __init__(
             "ibm/granite-20b-multilang-lab-rc",
             "ibm/granite-20b-multilingual",
             "ibm/granite-3b-code-plus-v1",
+            "ibm/granite-20b-code-javaenterprise",
```
fabianvf, line 81: I wonder if we should be validating them so strictly. Feel like it would make more sense to have a list of models we know work well, and issue a warning if someone is using a different model. Not on this PR but just a thought

fabianvf (approved):

kai/model_provider.py
jwmatthews: Totally agree, doesn't feel good to be this strict when the values are frequently changing.

jwmatthews: We have the below issues tracking to fix this

- Issue #129 
- Issue #164 (probably related)

PR #184, opened 2024-05-09
Fix max_workers comparison in run_demo.py

jmontleon:
If `KAI_MAX_WORKERS` is set I get an error from `run_demo.py`: `TypeError: '<=' not supported between instances of 'str' and 'int'`


JonahSussman (approved):
LGTM

PR #186, opened 2024-05-13
Openai compatible alts

jmontleon:


jwmatthews (approved):

jwmatthews:
@jmontleon good to merge this after the conflict is handled

jmontleon:
@jwmatthews sounds good. I'm going to add a note about OpenShift AI using vLLM as well, since it is also OpenAI compatible. I'm still trying to test, but it's mostly some pain copying models around as well as some pain with istio and certs.

tl;dr the way istio does routes is not friendly with the cert-manager route operator so it's time consuming doing things manually. and last time i tried to use langchain with self-signed or invalid certs didn't go well.

PR #188, opened 2024-05-13
Remove old kai-service directory

jmontleon:


JonahSussman (approved):
LGTM, I don't think we are either

PR #189, opened 2024-05-13
Add script to import reports from Konveyor hub

fabianvf:
Still TODO:
- [x] Use inventory API to retrieve credentials for private repos
- [x] Documentation
- [x] Tests

I am testing with konveyor deployed with this CR:
```yaml
apiVersion: tackle.konveyor.io/v1alpha1
kind: Tackle
metadata:
  name: tackle
  namespace: konveyor
spec:
  analyzer_fqin: quay.io/jortel/tackle2-addon-analyzer:kai-support
  feature_auth_required: "false"
  hub_image_fqin: quay.io/jortel/tackle2-hub:kai-support
```

To load the sample data into an unauthenticated konveyor instance, run 
```bash
$ ./samples/add_data_to_hub.py {your hub url} -k
```

To start the importer, run: 
```bash
$ make run-konveyor-importer HUB_URL={your hub url} IMPORTER_ARGS='--skip-verify'
```


kai/import_from_hub.py
```diff
@@ -0,0 +1,338 @@
...
+            raise ValueError("Key must be a string")
+        setattr(self, key, value)
+
+    def get(self, key, default=None):
+        if key in self:
+            return self[key]
+        return default
+
+
+class Incident(KaiBaseModel):
```
JonahSussman, line 43: Should we think about removing [analyzer_types.py](https://github.com/konveyor-ecosystem/kai/blob/main/kai/models/analyzer_types.py) if we have these instead? I think it'd be good to consolidate the models

kai/report.py
```diff
@@ -10,35 +10,61 @@
...
-        self.path_to_report = path_to_report
+    report = {}
+
+    def __init__(self, report_data: dict):
         self.workaround_counter_for_missing_ruleset_name = 0
-        ## We expect that self._read_report() will be the last statement
-        ## to invoke in this constructor
-        self.report = self._read_report()
+        self.report = self._reformat_report(report_data)
+
```
JonahSussman, line 18: This doesn't work intuitively. Consider this simplified case:

```python
class Report:
    report = {}

r1, r2 = Report(), Report()
r1.report['name'] = 'thing'
print(r2.report)  # Output: {'name': 'thing'}
```

Granted, self.report is set on line 17 so this is kinda mute, but just trying to avoid a gotcha. It might be better to just put a type hint like `report: dict`

kai/service/incident_store/incident_store.py
```diff
@@ -12,6 +12,29 @@
 from kai.models.kai_config import KaiConfigIncidentStore, KaiConfigIncidentStoreProvider
 from kai.report import Report
 
+# These prefixes are sometimes in front of the paths, strip them.
+# Also strip leading slashes since os.path.join can't join two absolute paths
+KNOWN_PREFIXES = ("/tmp/source-code/", "/addon/source/", "/")
```
JonahSussman, line 17: Consider changing this to 

```python
KNOWN_PREFIXES = ("/tmp/source-code/", "/addon/source/", "/")  # trunk-ignore(bandit/B108)
```
So Trunk is happy

kai/service/incident_store/psql.py
```diff
@@ -279,29 +284,31 @@ def load_report(self, app: Application, report: Report) -> tuple[int, int, int]:
...
-                    ),
-                )
+                # NOTE: When retrieving uris from the report, some of them
+                # had "/tmp/source-code/" as their root path. Unsure where
+                # it originates from.
+                file_path = remove_known_prefixes(unquote(urlparse(si[4]).path))
+                # file_path = os.path.join(
+                #     repo_path,
+                #     in_repo_path,
+                # )
```
JonahSussman, line 294: As long as `git diff` works with the in-repo path as opposed to the full path then I'm good with this. Just pointing out the potential functionality difference.

kai/import_from_hub.py
```diff
@@ -0,0 +1,338 @@
...
+def clone_repo_at_commit(repo_url, branch, commit, destination_folder):
+    try:
+        subprocess.run(
+            ["git", "clone", "-b", branch, repo_url, destination_folder], check=True
+        )
+    except subprocess.CalledProcessError as e:
+        KAI_LOG.error(f"An error occurred: {e}")
+
+    try:
+        subprocess.run(["git", "checkout", commit], cwd=destination_folder, check=True)
```
JonahSussman, line 301: Consider adding the `trunk-ignore` comments if these are spurious

kai/import_from_hub.py
```diff
@@ -0,0 +1,338 @@
+#!/usr/bin/env python
+
+import argparse
+import os
+import pprint
+import subprocess
```
JonahSussman, line 6: Consider adding the trunk-ignore comments if this is spurious

JonahSussman (changes requested):
Looks good, just a few minor nits

kai/import_from_hub.py
fabianvf: Yeah maybe, these models correspond to the konveyor API types though which are slightly different from the types in the report itself.

kai/import_from_hub.py
fabianvf: yeah a bunch of trunk comments still need to be added

kai/report.py
fabianvf: Good catch

kai/service/incident_store/psql.py
fabianvf: Yeah, I believe that the `Repo` object handles the relative path stuff

jmontleon:
I ran through the process this morning without problems other than environment issues addressed in separate documentation PR.

PR #198, opened 2024-05-21
Jinja Prompt Builder

JonahSussman:
Switched prompt builder to use jinja. Old prompt builder was unmaintainable (and I can say that cause I wrote it)


kai/server.py
```diff
@@ -284,6 +288,15 @@ def app():
     KAI_LOG.info(f"Selected provider: {config.models.provider}")
     KAI_LOG.info(f"Selected model: {webapp['model_provider'].model_id}")
 
+    webapp["jinja_env"] = Environment(
```
fabianvf, line 291: I think I'd put this farther down in the stack. It's more of an implementation detail within the llm_io_handler than something the server needs to be concerned about IMO

kai/llm_io_handler.py
```diff
@@ -70,6 +70,7 @@ def get_key_and_res_function(
 async def get_incident_solutions_for_file(
     model_provider: ModelProvider,
     incident_store: IncidentStore,
+    jinja_env: Environment,
```
fabianvf, line 73: I'd remove this as an argument and just have llm_io_handler build out the template. That way if we ever change the format we don't need to worry about plumbing it through as much

kai/models/kai_config.py
```diff
@@ -46,7 +46,8 @@ class KaiConfigIncidentStoreInMemory(BaseModel):
 class KaiConfigModels(BaseModel):
     provider: str
     args: dict
-    template: str = Field(default="")  # TODO: Hack until we get better templating
+    template: str = Field(default="main.jinja")
```
fabianvf, line 49: Do you think we could default to `{model-id}.jinja` with a fallback to main? That way we can have dedicated model-specific templates without swapping out configs.

kai/data/templates/main.jinja
fabianvf, line 1: We should also remove the obsoleted templates since they won't work anymore.

kai/server.py
JonahSussman: Sounds good, implementing

kai/llm_io_handler.py
```diff
@@ -18,6 +19,43 @@
...
+    jinja_env = Environment(**jinja_kwargs)  # trunk-ignore(bandit/B701)
+
+    if model_provider.template is not None:
+        return jinja_env.get_template(model_provider.template).render(pb_vars)
+
+    # Try to render the template corresponding to the model_id, fallback to
+    # main.jinja
+    try:
+        return jinja_env.get_template(model_provider.model_id).render(pb_vars)
+    except TemplateNotFound:
```
fabianvf, line 55: Maybe put a debug log in here, just in case someone expects a specific model template to be selected so that they can see it's not being found. Other than that looks good :+1:

PR #199, opened 2024-05-29
Fix template pathing and git repo requirement

JonahSussman:


kai/evaluation.py
```diff
@@ -216,16 +225,16 @@ def evaluate(
                 llm_result=llm_result.content,
             )
 
+            shutil.rmtree(os.path.join(full_example_path, ".git"))
```
fabianvf, line 228: We should only do this if we created the directory

JonahSussman: On line 166, I create the git repo if it doesn't exist already, so it should never fail unless I'm missing something

fabianvf: I'm not worried about it failing, I'm worried about it succeeding 😂

JonahSussman: My bad mistyped 😂

I think I'm not quite understanding. On line 166, I create a git repo if I don't detect a `.git` folder initially. Creating a git repo creates a `.git` folder. So there should always be a `.git` folder to delete on this line, correct? And I don't forsee us ever creating benchmarks that use the `.git` folder, so creating and deleting it like this is safe, I think

fabianvf: But what if it's pointing at a directory that's already a git repo?

JonahSussman: Gotcha, added fix

kai/evaluation.py
```diff
@@ -225,7 +228,8 @@ def evaluate(
                 llm_result=llm_result.content,
             )
 
-            shutil.rmtree(os.path.join(full_example_path, ".git"))
+            if created_git_repo:
```
fabianvf, line 231: This should probably be in the `finally` section of a `try` block, just so we know it triggers even if something breaks

JonahSussman: Good catch! Implemented

kai/llm_io_handler.py
```diff
@@ -30,11 +30,13 @@ def get_prompt(
     pb_vars: dict,
     path_templates: str = PATH_TEMPLATES,
     jinja_kwargs: dict = None,
+    should_fallback: bool = True,
```
fabianvf, line 33: I'd just go with `fallback`, the should feels redundant

kai/data/benchmarks/configs/default.toml
```diff
@@ -2,10 +2,11 @@
 
 [models]
 provider = "ChatIBMGenAI"
+# Can either be located in data/templates or the full path of the template
```
fabianvf, line 5: Maybe also a note pointing to the default

fabianvf (commented):
Sorry for the piecemeal review, should have done it all in one go. Few more thoughts but looks good to me

fabianvf (approved):

jmontleon (approved):

PR #202, opened 2024-06-10
Update install instructs

jmontleon:


jmontleon:
We're not installing kai in the python virtual environment.

Some things work despite that because we're doing tricks like this:
https://github.com/konveyor-ecosystem/kai/blob/main/example/run_demo.py#L13

But I recently had problems trying to run commands in @fabianvf's PR.

`-e` is just editable. It creates  env/lib/python3.12/site-packages/kai.egg-link that points to the dev directory so that if you update code or change branches it takes effect without going through a dance to reinstall.

PR #203, opened 2024-06-10
Add documentation of how stored incidents are used

fabianvf:


docs/solvedIncidentStore.md
```diff
@@ -0,0 +1,51 @@
+# Approach to Stored Analyses and Solved Incidents
+
+## Objective
+
+To ingest and permanently store all reports from Kai, using the data to determine and store solutions for fixed incidents, and to support future data extraction and fine-tuning.
```
jwmatthews, line 5: Should this be "store all reports from Konveyor"?

docs/solvedIncidentStore.md
```diff
@@ -0,0 +1,51 @@
...
+2. **Incident Fix Detection**
+3. **Diff Generation and Storage**
+4. **Solution Validation**
+5. **Long-term Data Storage**
+
+## Component Descriptions
+
+### Ingestion of Analysis Reports
+
+All analysis reports generated by Kai are ingested and stored permanently. This ensures that all data is retained for future reference and analysis.
```
jwmatthews, line 21: Similar to prior question, should this be Konveyor?

docs/solvedIncidentStore.md
```diff
@@ -0,0 +1,51 @@
...
+
+### Incident Fix Detection
+
+The recorded incidents in each analysis report, along with commit information, are used to determine whether an incident was fixed between two reports. This comparison helps identify changes and resolved incidents.
+
```
jwmatthews, line 26: Is it worth calling out how we determine an incident has been fixed and didn't just relocate to a different line number? ... even if we are not yet doing this, just calling out we know this kind of problem exists and is an area to improve could help.

i.e., I assume for an analysis it's likely that there is an unrelated code change so the lineNumber changes, yet its in fact the same incident, just relocated.

docs/solvedIncidentStore.md
fabianvf: whoops yes

jwmatthews (approved):

PR #211, opened 2024-06-27
Refactor PSQL incident store into ORM (SQLAlchemy)

JonahSussman:
You will have to purge the `data` volume, delete all containers using it.

```sh
podman rm $(podman ps -a --filter volume=data -q)
podman volume rm data
```

Should do the trick.


kai/service/incident_store/psql.py
```diff
@@ -1,93 +1,182 @@
...
 
+class PSQLUnmodifiedReport(Base):
+    __tablename__ = "unmodified_reports"
 
-# TODO(@JonahSussman): Migrate this to use an ORM
-class PSQLIncidentStore(IncidentStore):
-    def __init__(self, args: KaiConfigIncidentStorePostgreSQLArgs):
-        self.emb_provider = EmbeddingNone()
+    application_name: Mapped[str] = mapped_column(primary_key=True)
+    generated_at: Mapped[datetime.datetime] = mapped_column(
```
fabianvf, line 47: I think we probably don't want to use this as a primary key, since it will be unique if you insert the same report twice. I think ideally if we picked up the same report twice we would not duplicate it.

kai/service/incident_store/psql.py
```diff
@@ -1,93 +1,182 @@
...
 
-    # Abstract base class implementations
 
-    def delete_store(self):
-        with self.conn.cursor() as cur:
-            cur.execute(open(os.path.join(PATH_SQL, "drop_tables.sql"), "r").read())
+class PSQLRuleset(Base):
+    __tablename__ = "rulesets"
+
+    ruleset_name: Mapped[str] = mapped_column(primary_key=True)
```
fabianvf, line 78: Rulesets can change over time, not totally sure it matters though

fabianvf (commented):
Few questions but looks way nicer/more maintainable :+1:

kai/service/incident_store/psql.py
JonahSussman: My thought process was that this was just a record of all reports in the order they were processed. What are you thinking as the primary key?

fabianvf: I think that is accurate, but we also don't want to duplicate the reports coming in. I think ideally we can pull an identifier or combination of identifiers off the Report itself. I say this mostly because the Hub importer polls the API, but if it restarts it could suck up all the previous reports again, so we just want to make sure those only go in once (or we could see like a `crashloop but successfully load 50 reports` where the database just balloons in size). We could do that filter elsewhere as well if needed but if there's an identifier on the report that can be used I think that would be ideal

kai/service/incident_store/psql.py
fabianvf: oh right since this is just the violations, nvm ignore me. I was thinking if the rules change it would make a difference but nvm

kai/service/incident_store/psql.py
JonahSussman: Gotcha, that makes sense. How about storing the commit that generated the report instead?

fabianvf: hmm, I think that could be part of an identifier but we could feasibly get multiple reports against the same commit. I'd be cool adding an ID field to the report object and requiring us to pass it in. I think we'll have report IDs on the konveyor side so we'd just need to figure out how to populate it in the on-disk scenario

JonahSussman: Ok solid, I think I see what you're saying now

kai/server.py
```diff
@@ -273,13 +273,15 @@ def app():
...
 
-    webapp["incident_store"] = IncidentStore.from_config(config.incident_store)
-    KAI_LOG.info(f"Selected incident store: {config.incident_store.provider}")
-
     webapp["model_provider"] = ModelProvider(config.models)
     KAI_LOG.info(f"Selected provider: {config.models.provider}")
     KAI_LOG.info(f"Selected model: {webapp['model_provider'].model_id}")
 
+    webapp["incident_store"] = IncidentStore.from_config(
+        config.incident_store, webapp["model_provider"]
```
JonahSussman, line 281: Passing the model provider to the incident store so we can do LLM summaries next

kai/service/incident_store/incident_store.py
```diff
@@ -176,48 +206,404 @@ class Solution:
...
+
+    violation: Mapped[SQLViolation] = relationship(back_populates="incidents")
+    application: Mapped[SQLApplication] = relationship(back_populates="incidents")
+    solution: Mapped[SQLAcceptedSolution] = relationship(back_populates="incidents")
+
+    def __repr__(self) -> str:
+        return f"SQLIncident(violation_name={self.violation_name}, ruleset_name={self.ruleset_name}, application_name={self.application_name}, incident_uri={self.incident_uri}, incident_snip={self.incident_snip:.10}, incident_line={self.incident_line}, incident_variables={self.incident_variables}, solution_id={self.solution_id})"
+
+
+# def dump(sql, *multiparams, **params):
```
fabianvf, line 340: Can drop this I assume

fabianvf (approved):
Looks great!

PR #221, opened 2024-07-01
:sparkles: Add Containerfile

jmontleon:


build/entrypoint.sh
```diff
@@ -0,0 +1,26 @@
...
+  sed -i "s/^password =.*/password =\"$POSTGRES_PASSWORD\"/g" /kai/kai/config.toml
+fi
+
+TABLE=applications
+SQL_EXISTS=$(printf '\dt "%s"' "$TABLE")
+if [[ $(PGPASSWORD="$POSTGRES_PASSWORD" psql -h "kai_db" -U $POSTGRES_USER -d $POSTGRES_DB -c "$SQL_EXISTS") ]]
+then
+    echo "load-data has run already"
+else
+    echo "load-data has not run yet, starting ..."
```
fabianvf, line 18: Should we put a flag to skip load-data in case we want to pull from conveyor instead? Or we could do it in a follow up if we're just trying to containerize the standard demo real quick

fabianvf (approved):
One question but the changes look good

build/entrypoint.sh
jmontleon: Makes sense. I can set add an env var and add a conditional.

jwmatthews:
To build on Mac arm64 I needed these tweaks to requirements.txt:  

https://gist.github.com/jwmatthews/cf0fd9be2925337d8ddc8e0189528f31


Then saw success building with:

```
$ time podman build -f build/Containerfile . -t quay.io/konveyor/kai:latest
...
...
Successfully tagged quay.io/konveyor/kai:latest
e7b05979e2ede334ff60533be389033b59735650c77898e99b205f0d2b1de1f6
podman build -f build/Containerfile . -t quay.io/konveyor/kai:latest  16.49s user 3.39s system 14% cpu 2:18.41 total

```

jwmatthews (approved):

PR #222, opened 2024-07-01
:sparkles: Add podman / docker compose file

jmontleon:


jmontleon:
Probably need to add a quick explanation of how to use this in the README, but the gist of it is, edit kai/kai/config.toml as usual to set your provider and model then `podman compose up`.

If it's the first time running it will go through the data loading process, otherwise it will be skipped.

Also, as of yet there is no quay.io/konveyor/kai:latest image yet, so you'll have to build that for yourself for the moment. I'll get a GH action set up in the next day or so to remedy that.

compose.yaml
```diff
@@ -0,0 +1,39 @@
+version: "3.4"
+
+x-common-variables: &common-variables
+  POSTGRES_USER: foo
+  POSTGRES_PASSWORD: bar
+  POSTGRES_DB: baz
+
+services:
```
fabianvf, line 8: We should probably add the hub importer at some point, but could do it later if we're just trying to get the core demo running

fabianvf (approved):

jwmatthews (approved):
Tested on mac arm64 and confirmed working via run of run_demo.py against the podman composed containers.

PR #226, opened 2024-07-02
:ghost: Remove unnecessary line from Containerfile

jmontleon:


jmontleon:
This was from an early attempt at copying the config.toml in that I realized was unnecessarily complex. I missed deleting it when simplifying the process.

jwmatthews (approved):

PR #231, opened 2024-07-09
:sparkles: Container and Compose Improvements

jmontleon:


jmontleon:
It seems cloning over http(s) can be problematic with some ISPs. Since @sshaaf has been experiencing this it seems reasonable to believe others may as well, so I have moved the git clone to the container build so there is no need to clone at runtime.

I have also made it so the image and/or tag can be overridden. To do so replace either or both like so:
`IMAGE=quay.io/jmontleon/kai TAG=latest podman compose up`

fabianvf (approved):

PR #232, opened 2024-07-12
:sparkles:  Incident Infrastructure Improvements

JonahSussman:


kai/config.toml
```diff
@@ -4,28 +4,32 @@ log_dir = "$pwd/logs"
 demo_mode = false
 trace_enabled = true
 
-# **Postgresql incident store**
-# [incident_store]
-# provider = "postgresql"
+solution_consumers = ["diff_only", "llm_summary"]
```
fabianvf, line 7: Might be nice to either add a little detail on what it means here or add a link to some kind of explanation

kai/config.toml
```diff
@@ -4,28 +4,32 @@ log_dir = "$pwd/logs"
...
-# provider = "sqlite"
-#
 # [incident_store.args]
+# provider = "sqlite"
 # connection_string = "sqlite:///:memory:"
+# ```
 
 [incident_store]
-provider = "postgresql"
+solution_detectors = "naive"
```
fabianvf, line 27: same here

kai/data/templates/main.jinja
```diff
@@ -33,21 +33,10 @@ Source file contents:
...
-{{ incident.solved_example_file.before }}
-```
-Solved example after changes:
-```{{ incident.src_file_language }}
-{{ incident.solved_example_file.after }}
-``` #}
+incident to fix: "{{ incident.message }}"
+Line number: {{ incident.line_number }}
+{% if incident.solution_str is defined %}
+{{ incident.solution_str }}
```
fabianvf, line 39: Is there still a way for users to override the templates for the solutions?

kai/routes/change_model.py
```diff
@@ -0,0 +1,8 @@
+from aiohttp.web_request import Request
+
+from kai.routes.util import to_route
+
+
+@to_route("post", "/change_model")
+async def post_change_model(request: Request):
```
fabianvf, line 7: what is this for?

kai/routes/ws/get_incident_solution.py
```diff
@@ -0,0 +1,62 @@
...
+from typing import Iterator
+
+from aiohttp import web
+from aiohttp.web_request import Request
+from langchain_core.messages import BaseMessageChunk
+
+from kai.routes.util import to_route
+from kai.service.kai_application.kai_application import KaiApplication
+
+# TODO(@JonahSussman): Figure out proper pydantic model validation for this
```
fabianvf, line 11: also add a TODO about figuring out if we should remove this and the other get_incident_solution

kai/service/incident_store/__init__.py
```diff
@@ -1,11 +1,9 @@
 from .incident_store import Application, IncidentStore, Solution
-from .psql import PSQLIncidentStore
-from .sqlite import SQLiteIncidentStore
 
 __all__ = [
     "IncidentStore",
     "Solution",
-    "PSQLIncidentStore",
+    "PSQLIncidentStoreBackend",
```
fabianvf, line 6: What's the motivation for the name change here?

requirements.txt
```diff
@@ -18,18 +19,23 @@ attrs==23.1.0
 Babel==2.13.1
 backcall==0.2.0
 beautifulsoup4==4.12.2
+black==24.4.2
```
fabianvf, line 22: why was  this added in?

fabianvf (commented):
few questions, looks pretty good though and I like all the added tests. A lot of stuff was added to requirements.txt, just want to double check they are all required.

kai/routes/change_model.py
JonahSussman: Originally had this route if the user wanted to change the model via a POST request, but it's dead code as of now. Removing.

kai/data/templates/main.jinja
JonahSussman: The `get_prompt` still retrieves its prompt from the model provider, with a default of `main.jinja`, so that functionality is still there. Right now, there's no way to override the `solution_handling` templates though.

kai/service/incident_store/__init__.py
JonahSussman: I wanted it to be clear that the IncidentStore and postgres have a has-a relationship, instead of an is-a relationship. 

Though, I'm not sure if `__all__` is really needed in this `__init__.py` anymore.

requirements.txt
JonahSussman: Was doing some experiments with some other libraries. Looking into it.

kai/kai_trace.py
```diff
@@ -37,15 +36,14 @@ class Trace:
 
     def __init__(
         self,
-        config: KaiConfig,
```
jwmatthews, line 40: I think tracing might be broken, it used the config for 2 things

1) Check if tracing is enabled/disabled
2) Get the log directory

I assume tracing will get up to the spot below and look for self.config.log_dir and not find it.

https://github.com/konveyor/kai/blob/9bfb7be7d78350cd18ea93001098bb3b6e8d43de/kai/kai_trace.py#L55

jwmatthews: I'm seeing the below when I run

[2024-07-23 17:32:55 -0400] [87706] [ERROR] Error handling request
Traceback (most recent call last):
  File "/Users/jmatthews/git/jwmatthews/kai/env/lib/python3.12/site-packages/aiohttp/web_protocol.py", line 452, in _handle_request
    resp = await request_handler(request)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/jmatthews/git/jwmatthews/kai/env/lib/python3.12/site-packages/aiohttp/web_app.py", line 543, in _handle
    resp = await handler(request)
           ^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/jmatthews/git/jwmatthews/kai/kai/routes/get_incident_solutions_for_file.py", line 44, in post_get_incident_solutions_for_file
    trace = KaiTrace(
            ^^^^^^^^^
  File "/Users/jmatthews/git/jwmatthews/kai/kai/kai_trace.py", line 55, in __init__
    log_dir = process_log_dir_replacements(self.config.log_dir)
                                           ^^^^^^^^^^^
AttributeError: 'KaiTrace' object has no attribute 'config'

jwmatthews (changes requested):
I think we need to look over the rebase with:
 - Tracing
 - Logging to file

When I attempted to run I hit an exception with Tracing from a recent commit to let tracing follow the configured 'log' directory, looks like in the rebase we changed it up a little to not pass in the config object, but the code is still looking for the config object.

I think there is likely a piece missing for configuring the logging to file as well, when I ran I didn't see the 'logs' directory get created.

requirements.txt
JonahSussman: Looks like it was added by `datamodel-code-generator`, but I'm not using that anymore. Gonna delete and see if there are any issues.

kai/kai_trace.py
JonahSussman: Let's see if my most recent commit fixes this

jwmatthews (changes requested):
When I run 'run_demo.py' I see:

```
INFO - 2024-07-24 08:07:12,405 - __main__ - [         run_demo.py:101  -        write_to_disk()] - Writing updated source code to ./coolstore/src/main/java/com/redhat/coolstore/rest/RestApplication.java
ERROR - 2024-07-24 08:07:12,405 - __main__ - [         run_demo.py:106  -        write_to_disk()] - Failed to write updated_file @ ./coolstore/src/main/java/com/redhat/coolstore/rest/RestApplication.java with error: string indices must be integers, not 'str'
```

https://gist.githubusercontent.com/jwmatthews/ce27a70514fd0899cb852df78916cba8/raw/80b0a3d3dbf8d8b1eca7e3ae252ac6058c7ae65d/gistfile1.txt

kai/kai_trace.py
jwmatthews: I see logs written to disk and trace data written to disk.

We may need to tweak the Trace data.  Originally I had assumed that we'd create the Trace object once per request, currently we are creating it twice.  This functions, but the output is a little off where some of the layout is slightly off in regard to directory structure.  

This is a small tweak, the larger issue right now is that I can't run run_demo.py, it gets back a 200 response and has data that looks good, but the run_demo.py isn't able to parse the response to write it out.

Not sure what is wrong yet, almost looks like its not decoding to json and treating the response as a regular string

kai/data/templates/main.jinja
fabianvf: I wonder if we want to just pass variables in from the solution handler to the normal template. Could we declare what variables are declared by a handler and what variables are required for a template so that we can fail on start with an invalid config?

kai/kai_trace.py
JonahSussman: A `Response`s `.json()` method returns a `str` not a `dict`. Wondering why this wasn't an issue for me before... Pushed a fix, let's see if that helps.

requirements.txt
JonahSussman: I was able to remove the dependencies without issue. Gonna resolve this issue for now.

jwmatthews:
> When I run 'run_demo.py' I see:
> 
> ```
> INFO - 2024-07-24 08:07:12,405 - __main__ - [         run_demo.py:101  -        write_to_disk()] - Writing updated source code to ./coolstore/src/main/java/com/redhat/coolstore/rest/RestApplication.java
> ERROR - 2024-07-24 08:07:12,405 - __main__ - [         run_demo.py:106  -        write_to_disk()] - Failed to write updated_file @ ./coolstore/src/main/java/com/redhat/coolstore/rest/RestApplication.java with error: string indices must be integers, not 'str'
> ```
> 
> https://gist.githubusercontent.com/jwmatthews/ce27a70514fd0899cb852df78916cba8/raw/80b0a3d3dbf8d8b1eca7e3ae252ac6058c7ae65d/gistfile1.txt

run_demo.py is now working with @JonahSussman latest fix.
Verified run_demo.py is good with this PR on 2 different environments
- MacOS arm64 via a developer flow (running local)
- MacOS x86_64 via docker compose and local image build

jwmatthews (approved):
Verified success running with run_demo.py.

There is a slight tweak with the Trace data that could be addressed in this or a follow up PR

kai/data/templates/main.jinja
JonahSussman: I don't think I understand. 

`get_incident_solution_for_file` will pass the same variables to the user-specified prompt regardless. One of those variables, `incident.solution_str`, is generated by `solution_consumer`. Each solution consumer expects a `Solution`, which is indeed type-checked and uses prompts in the `solution_handling` directory. Right now, templates there can't be configured, but the main prompt can.

PR #237, opened 2024-07-18
:sparkles:  Allow logging to disk

jwmatthews:
This will:
- Introduce logging to a File
- Split out log levels for Console or File
- Fix podman compose logging so it writes to our local disk and not into the containers ephemeral storage
 
Misc:
 - Remove a lot of left over cruft from kai/__init_.py that is no longer used.
 - Quick fix for #239


fabianvf (approved):
LGTM

jmontleon (approved):
ACK

JonahSussman (approved):
LGTM

PR #249, opened 2024-07-23
👻 Log statement to show we are ready to receive requests

jwmatthews:
As I'm working up instructions for new users realized we don't have a good thing for users to read from logs to know when the server is ready to process a request.


JonahSussman (approved):
LGTM

PR #267, opened 2024-07-30
:bug: Closes #266 podman compose permission denied writing logs

### LINKED ISSUE #266: podman compose up results in a permission denied error writing logs on Fedora
When running `podman compose up` on Fedora the output is flooded with errors `PermissionError: [Errno 13] Permission denied: '/podman_compose/logs/kai_psql.log'`.

I opened a podman issue to try to get some insight into why we see a difference with podman(-compose) on Mac and Fedora.
https://github.com/containers/podman/issues/23439

I have also tried on a couple clean Fedora 40 VM installs with podman and podman-remote/machine and the issue reproduced, so I don't believe it is a system misconfiguration.

There are a few ways to proceed:
1. More manual steps:
```
mkdir logs
podman unshare chown 1001:1001 logs
podman compose up
```

2. Automated, but will cause problems with git:
Update mounts to `:rw,Z,U`

But this will break the git repo if you try to fetch updates or do anything else:
```
$ git fetch upstream
fatal: detected dubious ownership in repository at '/home/jason/Documents/go/src/github.com/konveyor/kai'
To add an exception for this directory, call:

	git config --global --add safe.directory /home/jason/Documents/go/src/github.com/konveyor/kai
```

3. Improving on 2:
- Populate a logs directory in the repo
- Add it to .gitignore
- Add an additional volume mount `      - ${PWD}/logs:/podman_compose/logs:rw,z,U`

I think unless we get some insight 3 is probably the most seamless approach.

jmontleon:


jwmatthews (approved):
Tested on MacOS and confirmed working.

I saw pydantic errors when I attempted to run ':stable' and ':latest' with this branch.
I downgraded kai/config.toml to what is in 'stable' and still saw pydantic issues

So to test, I built a new local image and used that local image with podman compose and verified this is working with run_demo.py

PR #268, opened 2024-07-30
:bug: fix run_demo.py imports

pranavgaikwad:
No idea how it was working for others. But in my environment, I had to bring this up to make sure all kai imports work

### DIFF
--- a/example/run_demo.py
+++ b/example/run_demo.py
@@ -11,15 +11,14 @@ from pathlib import Path
 
 import requests
 
-from kai.models.report_types import ExtendedIncident
-from kai.routes.get_incident_solutions_for_file import (
-    PostGetIncidentSolutionsForFileParams,
-)
-
 # Ensure that we have 'kai' in our import path
 sys.path.append("../../kai")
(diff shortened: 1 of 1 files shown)


jwmatthews (approved):
Others are following this: https://github.com/konveyor/kai/blob/main/docs/contrib/Dev_Environment.md#python-virtual-environment

So they are doing a: `pip install -e .`

I'm with you on being confused and adding the `sys.path.append("../../kai")` we had that in the script in an earlier incarnation and I think it got dropped with a refactor.

I can go either way with how we address.
I don't see harm with proceeding with this PR and adding the  `sys.path.append("../../kai")` 

@fabianvf @JonahSussman @jmontleon thoughts?

JonahSussman (changes requested):
Once the trunk error is fixed, it's good to merge. I have to run everything with  `PYTHONPATH=<project root> python ...` anyway. I don't know if there's a way to do this `sys.path.append` fix for everything though.

JonahSussman (approved):

PR #276, opened 2024-08-03
Added explicit steps for configuring and using custom rules.

### LINKED ISSUE #212: [Docs] Update demo.md to specifically call out how to invoke with our custom rules
We need to update the doc here:  https://github.com/konveyor-ecosystem/kai/blob/main/docs/demo.md

Currently it talks about running analysis but doesn't explicitly call out the need for the custom rules, to help others avoid questions, let's be explicit

This is how we run analysis:
https://github.com/konveyor-ecosystem/kai/blob/main/example/analyze.sh

We are using custom rules here:
https://github.com/konveyor-ecosystem/kai/tree/main/samples/custom_rules

Further... you will notice in https://github.com/konveyor-ecosystem/kai/blob/main/example/analyze.sh that we are also IGNORING the default rules baked in Kantra and overriding with https://github.com/konveyor-ecosystem/kai/tree/main/example/default_rules

The original intent of overriding what Kantra included by default was so we could tweak some of the included rules as we experimented.

As of June 2024, I am unsure we need to override the default rules.  I would say for terms of the demo doc, we just focus on the 'custom_rules' which are essential and provide needed pieces for our Java EE to Quarkus path.  

We can revisit to add the overridden default rules if needed, but as of today I am not sure this is required.

Myrausman:
Fixes #212

### DIFF
--- a/docs/customApps.md
+++ b/docs/customApps.md
@@ -74,7 +74,6 @@ Before running analysis, make sure that you have `kantra` installed
 1. `./get_latest_kantra_cli.sh` # fetches 'kantra' our analyzer tool and stores it in ../bin
 1. `cd ..`
 
-
 To ensure the analysis uses the necessary custom rules, follow these steps:
 
 - Download or create your custom rules and place them in the samples/custom_rules directory.
@@ -104,13 +103,14 @@ Analysis data will be stored in: `samples/analysis_reports/{APP_NAME}/<initial|s
 4. To fix issues identified during analysis using Kai, follow these [steps](https://github.com/konveyor-ecosystem/kai-vscode-plugin/blob/main/docs/user-guide.md#running-kai-fix)
 
 ---
+
 <b>Note: </b> There are two different places where analysis is performed and currently only one of them is set to ignore the default Kantra rules:
 
 1. Analysis of Sample Data: This runs on several sample applications using both default and custom rules. It simulates integration with a Konveyor instance populated with data. Configuration for this analysis is in [analyze_apps.py](https://github.com/konveyor/kai/blob/main/samples/analyze_apps.py#L83).
 
 2. Client-Side Analysis: This can be handled through example/run_demo.py or running in an IDE. Default rules are ignored, and only custom rules are used. Configuration for this analysis is in [analyze.sh](https://github.com/konveyor/kai/blob/main/example/analyze.sh#L50)
 
-For client-side analysis, use the ``` --enable-default-rulesets=false``` option and then pass in the below directory to use:
+For client-side analysis, use the ` --enable-default-rulesets=false` option and then pass in the below directory to use:
 [Konveyor Default Rules](https://github.com/konveyor/kai/tree/main/example/default_rules)
 
 ---
--- a/docs/customApps.md
+++ b/docs/customApps.md
@@ -79,7 +79,25 @@ To ensure the analysis uses the necessary custom rules, follow these steps:
 
 1. Custom Rules: Download or create your custom rules and place them in the samples/custom_rules directory.
 2. Ignore Default Rules: The analysis script is set to ignore default Kantra rules. You can review or adjust this in example/analyze.sh if needed.
+---
+<b>Note: </b> There are two different places where analysis is performed and currently only one of them is set to ignore the default Kantra rules:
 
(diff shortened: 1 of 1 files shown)


docs/customApps.md
```diff
@@ -74,6 +74,12 @@ Before running analysis, make sure that you have `kantra` installed
 1. `./get_latest_kantra_cli.sh` # fetches 'kantra' our analyzer tool and stores it in ../bin
 1. `cd ..`
 
+
+To ensure the analysis uses the necessary custom rules, follow these steps:
+
+1. Custom Rules: Download or create your custom rules and place them in the samples/custom_rules directory.
+2. Ignore Default Rules: The analysis script is set to ignore default Kantra rules. You can review or adjust this in example/analyze.sh if needed.
```
jwmatthews, line 81: Calling this out make sense, let's expand to include the 2 places that we perform analysis so folks are aware.

Note that we have 2 different places that perform analysis and currently only one of them is set to ignore the default Kantra rules.

I believe the intent of this document has focused on the analysis of sample data

1) Analysis of the 'sample data', this is analysis that runs on several sample applications where have initial/solved states.  This analysis is done to simulate what integration will look like when we run with a Konveyor instance that an organization has populated with data.   For this specific use-case we are running the below to analyze the sample apps:

https://github.com/konveyor/kai/blob/main/samples/analyze_apps.py#L83
`cmd = f'time ./bin/kantra analyze -i {source_dir} -t "quarkus" -t "jakarta-ee" -t "jakarta-ee8+" -t "jakarta-ee9+" -t "cloud-readiness" --rules {os.path.dirname(__file__)}/custom_rules -o {full_output_dir} --overwrite'`

In the above, we are allowing Kantra to use the default rules and we supplement with the custom rules.

2) We also have another place we run analysis, client side
This client side usage could be from the example/run_demo.py or from running in IDE.
Currently we are ignoring the default rules for the client side with example/run_demo.py and we experimented with a few small tweaks.

https://github.com/konveyor/kai/blob/main/example/analyze.sh#L50
`time "${KANTRA_BIN}" analyze -i "${PWD}"/"${SOURCE_DIR}" "${SOURCE_ONLY}" --enable-default-rulesets=false -t "quarkus" -t "jakarta-ee" -t "jakarta-ee8" -t "jakarta-ee9" -t "cloud-readiness" --rules "${DEFAULT_RULES_DIR}" --rules "${CUSTOM_RULES_DIR}" -o "${OUTDIR}" --overwrite`

We use the `--enable-default-rulesets=false` option client side and then pass in the below directory to use:
https://github.com/konveyor/kai/tree/main/example/default_rules

jwmatthews (changes requested):
Thank you for the contribution.

Please consider the comments I added to help improve the documentation and clear up confusion with the 2 places we run analysis.

Myrausman:
@jwmatthews , Can you please confirm if this is the content you meant to include in the document?

docs/customApps.md
```diff
@@ -74,6 +74,30 @@ Before running analysis, make sure that you have `kantra` installed
...
```
jwmatthews, line 81: @Myrausman this information you have is not technically correct, as the section you are documenting will NOT ignore default rules.  That is why I gave you more context so you could fully understand the 2 places we run analysis.

The documentation for this section:
https://github.com/konveyor/kai/blob/1ed90050c210c5a3597b0ef1a073eb41f4843008/docs/customApps.md#run-analysis

Is talking about how to run the analysis for the sample data which Kai will consult. 
That analysis DOES NOT ignore default rules.

Please remove:
`2. Ignore Default Rules: The analysis script is set to ignore default Kantra rules. You can review or adjust this in example/analyze.sh if needed.`

docs/customApps.md
```diff
@@ -74,6 +74,30 @@ Before running analysis, make sure that you have `kantra` installed
...
+---
+<b>Note: </b> There are two different places where analysis is performed and currently only one of them is set to ignore the default Kantra rules:
```
jwmatthews, line 83: Let's relocate this chunk of extra contextual information at the end of the document so we do not confuse folks.

jwmatthews (approved):
LGTM

jwmatthews (changes requested):
Almost ready, but need to adjust a `trunk fmt` issue

Please see https://github.com/konveyor/kai/blob/main/CONTRIBUTING.md#linting for how to install/run trunk to fix this.

jwmatthews (approved):

PR #255, opened 2024-07-24
:sparkles: Refactor requirements.txt and remove security vulnerabilities

JonahSussman:
Fixes #248.

In order to separate dependencies and dependencies-of-dependencies, I propose we switch to this following model of managing them:

**To install all dependencies in a new environment:**
- `python -m venv <venv-name>`
- `source <venv-name>/bin/activate`
- `pip install pip-tools`
- `pip-compile --allow-unsafe`
- `pip install -r requirements.txt`

**To add a new dependency to the project:**
- Add it to `requirements.in`
- `pip-compile --allow-unsafe`
- `pip install -r requirements.txt`

---

Below, I've also compiled a table of the dependencies I cleaned up in this PR:

| package | description | removed reason |
| --- | --- | --- |
| ansi2html | html color codes | not used |
| appnope | MacOS app nap prevention | not used |
| backcall | Specifications for callback functions passed in to an API | ipython no longer requires it |
| filelock | A platform independent file lock | Used by torch, huggingface-hub, triton, transformers, which are not used |
| fsspec | File system specification | Used by torch, huggingface-hub, which are not used |
| grpcio | HTTP/2-based RPC framework | Used by grpcio-reflection, which is not used |
| grpcio-reflection | Standard Protobuf Reflection Service for gRPC | Used by caikit-nlp-client, which is not used |
| huggingface-hub | Hugging Face model hub | Used by tokenizers, sentence-transformers, transformers, which are not used |
| InstructorEmbedding | Embedding for Instructor | not used |
| ipython-genutils | IPython utilities | ipython no longer requires it |
| joblib | Lightweight pipelining with Python functions | Used by nltk, which is not used |
| mpmath | Python library for arbitrary-precision floating-point arithmetic | Used by sympy, which is not used |
| networkx | Python package for creating and manipulating graphs and networks | Used by torch, which is not used |
| nltk | Natural Language Toolkit | Used by sentence-transformers, which is not used |
| nvidia-* | NVIDIA libraries | Used by torch, which is not used |
| pickleshare | Tiny 'shelve'-like database with concurrency support | ipython no longer requires it |
| pillow | Python Imaging Library | Used by torchvision, which is not used |
| protobuf | Protocol Buffers | Used by grpcio-reflection, which is not used |
| safetensors | SafeTensors | Used by transformers, which is not used |
| scikit-learn | Machine learning library | Used by caikit-nlp-client, which is not used | Used by sentence-transformers, which is not used |
[...]

PR #277, opened 2024-08-04
Update documentation for dependency management and setup instructions

### LINKED ISSUE #273: Update docs for how we manage python project dependencies
We need to update documentation as we recently changed how we manage python project dependencies

This could be a good first-issue for a new contributor who is versed in Python and is able to walk through the current install instructions, look at and understand #255, and then update the existing locations in our instructions.  I attempted to capture the 2 places I can think of below.

Update:

- Guidance on how to add modify a python dependency:  https://github.com/konveyor/kai/blob/main/CONTRIBUTING.md#updating-requirementstxt
- Guidance on installing dependencies in a virtual environment: https://github.com/konveyor/kai/blob/main/docs/contrib/Dev_Environment.md#local-development

with info from #255

example

**To install all dependencies in a new environment:**
- `python -m venv <venv-name>`
- `source <venv-name>/bin/activate`
- `pip install pip-tools`
- `pip-compile --allow-unsafe`
- `pip install -r requirements.txt`

**To add a new dependency to the project:**
- Add it to `requirements.in`
- `pip-compile --allow-unsafe`
- `pip install -r requirements.txt`

---

brf153:
Closes #273

Updated documentation to include new dependency management steps and virtual environment setup instructions as discussed in #255

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -190,18 +190,23 @@ have the right to contribute the code you are submitting to the project.
 You sign-off by adding the following to your commit messages. Your sign-off must
 match the git user and email associated with the commit.
 
+```bash
     This is my commit message
-
     Signed-off-by: Your Name <your.name@example.com>
+```
 
 Git has a `-s` command line option to do this automatically:
 
+```bash
     git commit -s -m 'This is my commit message'
+```
 
 If you forgot to do this and have not yet pushed your changes to the remote
 repository, you can amend your commit with the sign-off by running
 
+```bash
     git commit --amend -s
+```
 
 ## Pull Request Checklist
 
@@ -217,7 +222,7 @@ before you submit your code:
 - Ensure that unit tests pass
   - See [docs/contrib/Testing.md](docs/contrib/Testing.md)
 - If adding a new feature please add a new unit test
-- If you modified `requirements.txt` please see [updating requirements.txt](#updating-requirementstxt)
+- If you modified `requirements.txt` please see [Modifying a Python Dependency](#modifying-a-python-dependency)
 - Ensure that [`example/run_demo.py`](example/run_demo.py) works
 - Commits are signed as per [DCO](#dco)
 - PR Title begins with a gitemoji as described in [Pull Request Title](#pull-request-title)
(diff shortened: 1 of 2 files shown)


CONTRIBUTING.md
```diff
@@ -132,6 +132,37 @@ See [docs/contrib/Dev_Environment.md](docs/contrib/Dev_Environment.md)
 
 - If you are a developer working on Kai and you are updating requirements.txt, you will need to do some manual changes beyond just a `pip freeze &> ./requirements.txt`, we have a few directives that address differences in 'darwin' systems that need to be preserved. These need to be added manually after a 'freeze' as the freeze command is not aware of what exists in requirements.txt. Please consult the diff of changes you are making now from prior version and note the extra directions for `python_version` and or `sys_platform`
```
jwmatthews, line 133: This section on the requirements.txt is no longer relevant and should be removed as we now follow the procedure of placing requirements in requirements.in and generating the requirements.txt with `pip-compile`

CONTRIBUTING.md
```diff
@@ -132,6 +132,37 @@ See [docs/contrib/Dev_Environment.md](docs/contrib/Dev_Environment.md)
...
+   ```bash
+      python -m venv <venv-name>
+      source <venv-name>/bin/activate
+      pip install pip-tools
+      pip-compile --allow-unsafe
+      pip install -r requirements.txt
+   ```
+  
+3. **Running the project:**
+    The following commands are used to the run the project
```
jwmatthews, line 154: Note that run_demo.py is only one way to potentially interact with Kai.  This is a great way for developers to get a feel and run a demo to check workflow, but the normal path for end users will be via the IDE plugin.  We can still call this out, but should phrase it as a recommendation of one possible way

CONTRIBUTING.md
```diff
@@ -132,6 +132,37 @@ See [docs/contrib/Dev_Environment.md](docs/contrib/Dev_Environment.md)
...
+
+    ```bash
+    make load-data
+    make run-server
+    python run_demo.py
```
jwmatthews, line 159: I assume this won't work as is since the directories are different.
The Makefile is in top level kai clone and the run_demo.py is in `example/run_demo.py`.

If we want to include this as guidance for one way to run if things are working, then let's update so the directory is explicit.

brf153:
@jwmatthews, thank you for reviewing the pull request. I have made the necessary changes as suggested.

jwmatthews:
Thank you @brf153

jwmatthews (approved):

brf153:
@jwmatthews I noticed there are some linting issues, but it seems they aren't related to the code I added. Would you like me to fix the linting issues and add another commit?

jwmatthews:
> @jwmatthews I noticed there are some linting issues, but it seems they aren't related to the code I added. Would you like me to fix the linting issues and add another commit?

Yes please, that would be helpful.

brf153:
@jwmatthews done

PR #285, opened 2024-08-09
📖 README updates so folks may run stable as of today

jwmatthews:
These are temporary doc changes we need to unblock any new folks who are interested in running Kai


savitharaghunathan (approved):
visiack

PR #289, opened 2024-08-09
✨ Ability to use Amazon Bedrock models such as Claude 3.5 Sonnet

jwmatthews:
This allows us to use models deployed in Amazon Bedrock

The Claude3 models seem to be a little behind with langchain integration, I needed to use an older library to access them as they aren't yet available in `langchain_aws.llms.bedrock.BedrockLLM`


requirements.txt
```diff
@@ -97,8 +107,6 @@ gitdb==4.0.11
     # via gitpython
 gitpython==3.1.43
     # via -r requirements.in
-greenlet==3.0.3
```
JonahSussman, line 100: Wonder why this is no longer necessary

JonahSussman (approved):
LGTM, just a small comment on the dependencies

requirements.txt
jwmatthews: I'm unsure why greenlet was removed, I don't think its from this PR.

I ran `pip-compile` on current main (not this PR) and I see greenlit is being removed

```
$ pip-compile --allow-unsafe

$ git diff
diff --git a/requirements.txt b/requirements.txt
index 1cb26bc..683b395 100644
--- a/requirements.txt
+++ b/requirements.txt
@@ -20,6 +20,8 @@ anyio==4.4.0
     #   httpx
     #   jupyter-server
     #   openai
+appnope==0.1.4
+    # via ipykernel
 argon2-cffi==23.1.0
     # via jupyter-server
 argon2-cffi-bindings==21.2.0
@@ -64,6 +66,8 @@ click==8.1.7
     #   typer
 comm==0.2.2
     # via ipykernel
+coverage==7.6.0
+    # via -r requirements.in
 dataclasses-json==0.6.7
     # via langchain-community
 debugpy==1.8.2
@@ -97,8 +101,6 @@ gitdb==4.0.11
     # via gitpython
 gitpython==3.1.43
     # via -r requirements.in
-greenlet==3.0.3
-    # via sqlalchemy
 gunicorn==22.0.0
     # via -r requirements.in
 h11==0.14.0
```

JonahSussman: As long as everything still works I'm good with it, just pointing it out

jmontleon: https://docs.sqlalchemy.org/en/20/faq/installation.html

```The greenlet dependency does not install by default for CPU architectures for which greenlet does not supply a [pre-built binary wheel](https://pypi.org/project/greenlet/#files). Notably, this includes Apple M1. To install including greenlet, add the asyncio [setuptools extra](https://packaging.python.org/en/latest/tutorials/installing-packages/#installing-setuptools-extras) to the pip install command:```

jmontleon: Pretty sure the above explains it.

jmontleon: If I update sqlalchemy to 2.0.32 in the requirements.in it looks like it is recognized as optional and unnecessary on Fedora as well.

jmontleon: And 2.0.32 should be fine to satisfy langchain and langchain-community:

https://github.com/langchain-ai/langchain/blob/langchain%3D%3D0.2.11/libs/langchain/pyproject.toml#L40

https://github.com/langchain-ai/langchain/blob/langchain-community%3D%3D0.2.10/libs/community/pyproject.toml#L35

jmontleon (approved):

PR #294, opened 2024-08-13
Coolstore evaluation

pranavgaikwad:
This is a series of experiments we performed for Coolstore application evaluating the responses we get from Kai for various different migration issues under different conditions. We use an LLM to evaluate responses themselves.


jwmatthews (approved):

notebooks/evaluation/group_rules_by_effort.py
```diff
@@ -0,0 +1,33 @@
...
+
+    
+    for effort, objects in grouped_data.items():
+        output_file = os.path.join(odir, f"{effort}.yaml")
+        print(f"dumping to {output_file}")
+        with open(output_file, 'w') as file: yaml.dump(objects, file)
+
+if __name__ == "__main__":
+    # replace following path with rules in your path
+    input_directory = '/home/pranav/Projects/rulesets/default/generated/quarkus/'
```
fabianvf, line 25: Probably makes sense to make this a CLI argument or envvar

pranavgaikwad: ended up not using that file at all, removed it now

fabianvf:
Would be awesome to add the writeup in the PR/at the end of the notebook as a doc as a followup, but doesn't need to happen now

PR #300, opened 2024-08-15
:sparkles: Add Line Match algorithm and Fix Various Bugs

JonahSussman:
Using https://github.com/JonahSussman/sequoia-diff to get tree mappings between ASTs.

Implements #298

### DIFF
--- a/kai/service/llm_interfacing/model_provider.py
+++ b/kai/service/llm_interfacing/model_provider.py
@@ -1,5 +1,4 @@
 import os
-from typing import Any
 
 from genai import Client, Credentials
 from genai.extensions.langchain.chat_llm import LangChainChatInterface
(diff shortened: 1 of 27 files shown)


kai/service/llm_interfacing/model_provider.py
```diff
@@ -137,3 +153,40 @@ def __init__(self, config: KaiConfigModels):
             ]
         else:
             self.llama_header = config.llama_header
+
+
+class Wrapper:
```
fabianvf, line 158: where is this being used?

fabianvf (approved):
One question but it looks sane to me

kai/service/llm_interfacing/model_provider.py
JonahSussman: I wanted to see how the ModelProvider's LLM was used during debugging. I'll move this to the test module.

PR #308, opened 2024-08-19
:ghost: Update compose DB image

jmontleon:
We are not making use of vector, so switching to the same image that we use in konveyor upstream. This is also similar to the UBI image we use downstream.

https://github.com/konveyor/operator/blob/main/bundle/manifests/konveyor-operator.clusterserviceversion.yaml#L280C26-L280C65

### DIFF
--- a/build/entrypoint.sh
+++ b/build/entrypoint.sh
@@ -3,13 +3,13 @@
 # If we are using podman or docker compose use the repo config.toml
 if [[ -f /podman_compose/kai/config.toml ]]; then
 	cp /podman_compose/kai/config.toml /kai/kai/config.toml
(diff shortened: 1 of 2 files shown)


build/entrypoint.sh
```diff
@@ -3,13 +3,13 @@
 # If we are using podman or docker compose use the repo config.toml
 if [[ -f /podman_compose/kai/config.toml ]]; then
 	cp /podman_compose/kai/config.toml /kai/kai/config.toml
-	sed -i "s/^host =.*/host = \"${POSTGRES_HOST}\"/g" /kai/kai/config.toml
```
fabianvf, line 6: note to self i need to update these on https://github.com/konveyor/operator/pull/374

fabianvf (approved):

fabianvf (approved):

Issue #213, opened 2024-06-27
We are urlencoding the source file contents before we send to LLM


jwmatthews:
We are starting to escape some portions of the source code we send to the LLM request.
Example:
 - `public List&lt;ShoppingCartItem&gt; getShoppingCartItemList() {`
 - `return &#34;ShoppingCart [cartItemTotal=&#34; + cartItemTotal`
 

Below is an example of the source file contents in a prompt we constructed
```
Source file contents:
```java
package com.redhat.coolstore.model;

import java.io.Serializable;
import java.util.ArrayList;
import java.util.List;

import javax.enterprise.context.Dependent;

@Dependent
public class ShoppingCart implements Serializable {

	private static final long serialVersionUID = -1108043957592113528L;

	private double cartItemTotal;

	private double cartItemPromoSavings;
	
	private double shippingTotal;
	
	private double shippingPromoSavings;
	
	private double cartTotal;
			
	private List&lt;ShoppingCartItem&gt; shoppingCartItemList = new ArrayList&lt;ShoppingCartItem&gt;();

	public ShoppingCart() {
		
	}
	
	public List&lt;ShoppingCartItem&gt; getShoppingCartItemList() {
		return shoppingCartItemList;
	}

	public void setShoppingCartItemList(List&lt;ShoppingCartItem&gt; shoppingCartItemList) {
		this.shoppingCartItemList = shoppingCartItemList;
	}

	public void resetShoppingCartItemList() {
		shoppingCartItemList = new ArrayList&lt;ShoppingCartItem&gt;();
	}

	public void addShoppingCartItem(ShoppingCartItem sci) {
		
		if ( sci != null ) {
			
			shoppingCartItemList.add(sci);
			
		}
		
	}
	
	public boolean removeShoppingCartItem(ShoppingCartItem sci) {
		
		boolean removed = false;
		
		if ( sci != null ) {
			
			removed = shoppingCartItemList.remove(sci);
			
		}
		
		return removed;
		
	}

	public double getCartItemTotal() {
		return cartItemTotal;
	}

	public void setCartItemTotal(double cartItemTotal) {
		this.cartItemTotal = cartItemTotal;
	}

	public double getShippingTotal() {
		return shippingTotal;
	}

	public void setShippingTotal(double shippingTotal) {
		this.shippingTotal = shippingTotal;
	}

	public double getCartTotal() {
		return cartTotal;
	}

	public void setCartTotal(double cartTotal) {
		this.cartTotal = cartTotal;
[...]

PR #317, opened 2024-08-22
🐛 Address URL encoding for source code sent to LLM 

SaxenaAnushka102:
Addresses #213 by applying the `| urlencode` filter to the src_file_contents variable in main.jinja. 

### Testing:
Ran `example/run_demo.py` with the modified main.jinja file to verify that the LLM is receiving the source code correctly and processing it as expected.

Signed-off-by: Anushka Saxena <anushkasaxena.1947@gmail.com>

### DIFF
--- a/kai/data/templates/main.jinja
+++ b/kai/data/templates/main.jinja
@@ -26,7 +26,7 @@ After you have shared your step by step thinking, provide a full output of the u
 File name: "{{ src_file_name }}"
 Source file contents:
 ```{{ src_file_language }}
-{{ src_file_contents | urlencode }} ##use jinja filter urlencode 
+{{ src_file_contents | safe }} 
 ```
 
 ## Issues
--- a/kai/data/templates/main.jinja
+++ b/kai/data/templates/main.jinja
@@ -26,7 +26,7 @@ After you have shared your step by step thinking, provide a full output of the u
 File name: "{{ src_file_name }}"
 Source file contents:
 ```{{ src_file_language }}
-{{ src_file_contents }}
+{{ src_file_contents | urlencode }} ##use jinja filter urlencode 
 ```
 
 ## Issues


kai/data/templates/main.jinja
```diff
@@ -26,7 +26,7 @@ After you have shared your step by step thinking, provide a full output of the u
 File name: "{{ src_file_name }}"
 Source file contents:
 ```{{ src_file_language }}
-{{ src_file_contents }}
+{{ src_file_contents | urlencode }} ##use jinja filter urlencode
```
jwmatthews, line 29: I believe this is the opposite behavior of what we want.  

As per #213, the issue is that sometimes we are mistakenly urlencoding parts of the source code which we do not want to do.  I am not certain why/where the urlencoding is coming in, but the issue is that we are doing this somewhere in limited areas and we do not want to do that.

For example, instead of outputting: `public List<ShoppingCartItem>` 
we yield `public List&lt;ShoppingCartItem&gt;`


I mentioned in https://github.com/konveyor/kai/issues/213#issuecomment-2195111637 
I used `{{ src_file_contents | safe }}' 

Ideally we'd localize and understand why is it that only some parts of the files are being encoded, yet for short-term it could be sufficient to just do the ` | safe` that seems to address the problem

jwmatthews (changes requested):
I added a comment explaining this is not the behavior we want to address #213

kai/data/templates/main.jinja
SaxenaAnushka102: Hello @jwmatthews! Apologies for the confusion. I misunderstood the specific encoding behavior we wanted to avoid, and I mistakenly added the urlencode filter, thinking we're not automatically escaping mentioned cases (<,>). 
Also, I'm not sure why the branch isn't reflected I'm merging from main :(
Please let me know if I shall raise another PR for this. Thanks!

jwmatthews:
@SaxenaAnushka102 I apologize that I made a mistake with order of merging a few PRs and this fix got included via #318.

I had been working on a client in #318, as I tested that I found a few other locations we needed to adjust the encoding.  I addressed what I saw during that PR's testing and forgot you had this PR open with the 1-line fix.  In hindsight I would have merged your PR first, I am sorry.


Thank you for your contribution and I apologize on the confusion with not merging your fix first.

PR #312, opened 2024-08-21
:sparkles: Adds a CLI frontend for kai

fabianvf:
Playing around with making kai a client-side tool rather than server-side tool. Just a quick hack.

```
$ kai/cli.py -h
usage: cli.py [-h] [-log {debug,info,warning,error,critical}] [-d | --demo-mode | --no-demo-mode] [-a APPLICATION_NAME] [-c CONFIG] [-r REPORT_PATH] [-i INCIDENTS] path

Get incident solutions for a file.

positional arguments:
  path                  The path to the file to be migrated

options:
  -h, --help            show this help message and exit
  -log {debug,info,warning,error,critical}, --loglevel {debug,info,warning,error,critical}
                        Provide logging level. Options: - debug: Detailed information, typically of interest only when diagnosing problems. - info: Confirmation that things are working as expected. - warning: An indication that something unexpected happened, or indicative of some problem in the near future (e.g.,
                        ‘disk space low’). The software is still working as expected. - error: Due to a more serious problem, the software has not been able to perform some function. - critical: A serious error, indicating that the program itself may be unable to continue running. Example: --loglevel debug
                        (default: warning)
  -d, --demo-mode, --no-demo-mode
  -a APPLICATION_NAME, --application-name APPLICATION_NAME
                        The name of the application
  -c CONFIG, --config CONFIG
                        The path to the config file
  -r REPORT_PATH, --report-path REPORT_PATH
                        Path to an analysis report. One of report-path or incidents are required
  -i INCIDENTS, --incidents INCIDENTS
                        JSON-formatted list of incidents. One of report-path or incidents are required
```

Example in use:

```
$ kai/cli.py --demo-mode --loglevel warning -c kai/config.toml -r example/analysis/coolstore/output.yaml example/coolstore/src/main/java/com/redhat/coolstore/rest/CartEndpoint.java
Console logging for 'kai' is set to level 'WARNING'
File logging for 'kai' is set to level 'DEBUG' writing to file: '/home/fabian/projects/github.com/konveyor-ecosystem/kai/logs/kai_server.log'
// Add import statements for jakarta.enterprise, jakarta.inject, and jakarta.ws.rs.
import jakarta.enterprise.context.SessionScoped;
import jakarta.inject.Inject;
import jakarta.ws.rs.DELETE;
import jakarta.ws.rs.GET;
import jakarta.ws.rs.POST;
import jakarta.ws.rs.Path;
import jakarta.ws.rs.PathParam;
import jakarta.ws.rs.Produces;
[...]

PR #318, opened 2024-08-23
Add an example client to move code fix generation client side

jwmatthews:
Extended what was started in #312 in an experiment to redo run_demo.py without solved incidents.

Idea is to move to the generate fix logic to client side, then add an agent in to take the result and vet it and work with a LLM to improve it.

At the moment this PR has the capability of run_demo.py, it does not yet have an agent in it.


JonahSussman (approved):
LGTM. Going to start CLI work once I'm done cleaning up `podman compose` stuff

pranavgaikwad (approved):

Issue #264, opened 2024-07-29
Mismatch of config.toml with git checkout and image in podman compose:   pydantic_core._pydantic_core.ValidationError: 1 validation error for KaiConfig -  Unable to extract tag using discriminator 'provider' [type=union_tag_not_found, input_value={'solution_detectors': 'n...'password': 'dog8code'}}, input_type=dict]


jwmatthews:
I think the below is due to a mismatch where we are tagging images to isolate all changes, but we allow the config to be modified and used from the local checkout.

So in my situation:'
 -> running `podman compose up` that uses a `:stable` tag pointing to:  [commit 65a5761d479d84c15ccf0769e2bf00ea71a4836c](https://github.com/konveyor/kai/commit/65a5761d479d84c15ccf0769e2bf00ea71a4836c)
 -> My kai checkout of 'main' is at [commit 11a28d32992d9888234c2ea3a630a52733a48842](https://github.com/konveyor/kai/commit/11a28d32992d9888234c2ea3a630a52733a48842)

With this combination, the code running inside of the :stable image is reading the local checkout to parse the kai/config.toml and its crashing as I suspect we changed structure of kai/config.toml

Error from `podman compose up`

```
kai-1     |   File "/kai/kai/server.py", line 265, in app
kai-1     |     config = KaiConfig.model_validate_filepath(
kai-1     |              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
kai-1     |   File "/kai/kai/models/kai_config.py", line 125, in model_validate_filepath
kai-1     |     return KaiConfig.model_validate(model_dict)
kai-1     |            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
kai-1     |   File "/opt/app-root/lib64/python3.12/site-packages/pydantic/main.py", line 551, in model_validate
kai-1     |     return cls.__pydantic_validator__.validate_python(
kai-1     |            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
kai-1     | pydantic_core._pydantic_core.ValidationError: 1 validation error for KaiConfig
kai-1     | incident_store
kai-1     |    pydantic_core._pydantic_core.ValidationError: 1 validation error for KaiConfig
kai-1     |     For further information visit https://errors.pydantic.dev/2.7/v/union_tag_not_found
```

https://gist.github.com/jwmatthews/2619140e479c4029b5e90d9639d41a68

[...]

PR #320, opened 2024-08-29
:sparkles: Clean up and simplify config management

JonahSussman:
Config is now sourced from the environment as well as the config.toml via [pydantic-settings](https://docs.pydantic.dev/latest/concepts/pydantic_settings/). This makes Kai align more with [12-factor config principles](https://12factor.net/config).

pydantic-settings allows for easy config overrides (higher overrides lower):
 - Config file that is declared on the command line / via init arguments. If using `podman compose up`, the config file in `build/config.toml` is supplied as a command line argument.
- Environment variables
- Local config file, `kai/config_local.toml`. This is for quick changes when running on bare metal.
- Global config file, `kai/config_default.toml`. Sane defaults.
- Default field values

**Task list:**
- [x] Integrate pydantic-settings into the main project
- [x] (Optional) Find a better solution to the config overriding thing
- [ ] Update documentation


README.md
```diff
@@ -87,7 +87,7 @@ The above information is a quick path to enable running Kai quickly to see how i
 ## Roadmap and Early Builds
 
 - Kai is in it's early development phase and is NOT ready for production usage.
-- See [ROADMAP.md](/ROADMAP.md) to learn about the project's goals and milestones
+- See [docs/ROADMAP.md](docs/ROADMAP.md) to learn about the project's goals and milestones
```
jwmatthews, line 90: I'm unsure about this relocation of ROADMAP.md.   I lean towards thinking keeping it at root directory of project is good so its very explicit for anyone coming into the project. 

I am thinking we'd want to keep these files at root level;  README.md, ROADMAP.md, CONTRIBUTING.md

build/Containerfile
```diff
@@ -5,11 +5,25 @@ RUN echo -e "[almalinux9-appstream]" \
  "\nbaseurl = https://repo.almalinux.org/almalinux/9/AppStream/\$basearch/os/" \
  "\nenabled = 1" \
  "\ngpgcheck = 0" > /etc/yum.repos.d/almalinux.repo
+
 RUN dnf -y install postgresql && dnf clean all
 USER 1001
-COPY --chown=1001:1001 . /kai
```
jwmatthews, line 10: @JonahSussman did you run into an issue that prompted the change to break out what is explicitly copied?

compose.yaml
```diff
@@ -1,47 +1,56 @@
...
+      # host system.
       GENAI_KEY:
       OPENAI_API_BASE:
       OPENAI_API_KEY:
     image: ${IMAGE:-quay.io/konveyor/kai}:${TAG:-stable}
     volumes:
-      - ${PWD}:/podman_compose:rw,z
-      - ${PWD}/logs:/podman_compose/logs:rw,z,U
+      - ${PWD}:/podman_compose:rw,Z
+      # - ${PWD}/logs:/podman_compose/logs:rw,Z,U
```
jwmatthews, line 41: Is this intended to comment out the logs directory for the 'kai' container?

Jason saw problems with SELinux in past and podman compose writing to the directory on the host, the workaround was to have the logs directory exist ahead of time and then mount it as a volume.  

On Mac, we did not see an issue, worked fine with just the higher level '/podman_compose' directory.

docs/contrib/Demo_Mode.md
```diff
@@ -4,35 +4,35 @@ The kai server is able to cache results from a LLM and later replay these to aid
...
+If you are running with KAI**DEMO_MODE=TRUE and you see a `401` being returned from the LLM Provider consider that this is likely 'correct' behavior, if you do not have a valid API key defined. Note that KAI**DEMO_MODE will attempt to play back a cached response, yet if there is no cached data for the request Kai will attempt to talk to the LLM and will make a 'real' request, which if you don't have a valid API key will result in a `401`. To 'fix' this you will want to look at the request you are sending to Kai and ensure it is cached for the model you have configured, if you don't have valid cached data then you will need to get a valid api key and re-run so the data may be cached.
```
jwmatthews, line 35: Is `KAI**DEMO_MODE=TRUE` intended?

docs/contrib/Dev_Environment.md
```diff
@@ -3,8 +3,8 @@
 ## Access to a Large Language Model (LLM)
 
 - If you want to run Kai against a LLM you will likely need to configure a LLM API Key to access your service (unless running against a local model)
-  - We do provide a means of running Kai against previously cached data from a few models to aid demo flows. This allows you to run through the steps of using previously cached data without requiring access to a LLM. Note, if you do not provide LLM API access then the DEMO_MODE flow will only be able to replay previous cached responses.
-    - We call this 'DEMO_MODE', i.e. `DEMO_MODE=true make run-server`
+  - We do provide a means of running Kai against previously cached data from a few models to aid demo flows. This allows you to run through the steps of using previously cached data without requiring access to a LLM. Note, if you do not provide LLM API access then the KAI\_\_DEMO_MODE flow will only be able to replay previous cached responses.
```
jwmatthews, line 6: Intended?  `KAI\_\_DEMO_MODE`

docs/contrib/Dev_Environment.md
```diff
@@ -3,8 +3,8 @@
...
+    - We call this 'KAI**DEMO_MODE', i.e. `KAI**DEMO_MODE=true make run-server`
```
jwmatthews, line 7: Intended? `KAI**DEMO_MODE'`

pyproject.toml
```diff
@@ -0,0 +1,78 @@
+[build-system]
+requires = ["setuptools >= 61.0"]
+build-backend = "setuptools.build_meta"
+
+[project]
+name = "kai"
+version = "0.0.1"
+dependencies = [
```
jwmatthews, line 8: Is this replacing requirements.in?

If so, let's also capture how to update a new dependency for contributors in
https://github.com/konveyor/kai/blob/main/CONTRIBUTING.md#modifying-a-python-dependency

build/entrypoint.sh
```diff
@@ -1,15 +1,12 @@
...
-if [[ -f /podman_compose/kai/config.toml ]]; then
-	cp /podman_compose/kai/config.toml /kai/kai/config.toml
-	sed -i "s/^host =.*/host = \"${POSTGRESQL_HOST}\"/g" /kai/kai/config.toml
-	sed -i "s/^database =.*/database = \"${POSTGRESQL_DATABASE}\"/g" /kai/kai/config.toml
-	sed -i "s/^user =.*/user = \"${POSTGRESQL_USER}\"/g" /kai/kai/config.toml
-	sed -i "s/^password =.*/password =\"${POSTGRESQL_PASSWORD}\"/g" /kai/kai/config.toml
+# If a custom config is specified, use it
+if [[ -f /podman_compose/build/config.toml ]]; then
+	printf "Using custom config.toml\n"
+	cp /podman_compose/build/config.toml /kai/kai/config.toml
```
jwmatthews, line 6: Will an end user running `podman compose up` have access to the `config.toml` on their local disk so they can see the settings and adjust?

The usecase we want to satisfy is to allow the end user to:
- View the `config.toml` from local disk
- Change settings on local disk to modify configured models

I'm not sure from a quick review of this PR if there is a mechanism for the end user to view/edit the `config.toml` that is used.  My current impression is we are copying it into the container at build time.  I see how that helps for running with the default configurations and I see the intent to override via environment variables, but unsure about how a user can modify the config.toml.

jwmatthews (commented):
Looking good overall, but I did have a few questions.

- Lets verify end user can see/edit the `config.toml` used, concern mostly with workflow of how an end user sees the configured model and may change this model.
- Let's consider keeping ROADMAP.md at the main project root.
- I see the logs directory is no longer part of compose.yaml, unsure if that is intended.

pyproject.toml
JonahSussman: Yes. Documentation is up next

build/entrypoint.sh
JonahSussman: The user can easily adjust their `config.toml`. The way it works in this PR is:

1. When the container is _built_, the `config.toml` that is present in the repo is copied over to the container
2. When the container is _run_, it first checks if the file `build/config.toml` exists. If it does, it copies it over to the container.

Thus, the user can specify their config via `build/config.toml`. There is an `example_config.toml` also present in that directory. We could even add `example_stable_config.toml` or something so the user has a config that can match their version. We can change the location that `entrypoint.sh` looks for if it makes more sense for the user to specify their custom config elsewhere.

docs/contrib/Demo_Mode.md
JonahSussman: No, I had an error with find and replace. Fixing.

compose.yaml
JonahSussman: This was a mistake, reverting. I wasn't seeing issues with SELinux, but I was seeing issues with `z` versus `Z`.

JonahSussman:
I updated how config priority is done. It's now done in the following way (higher overrides lower):
 - Config file that is declared on the command line / via init arguments. If using `podman compose up`, the config file in `build/config.toml` is supplied as a command line argument.
- Environment variables
- Local config file, `kai/config_local.toml`. This is for quick changes when running on bare metal.
- Global config file, `kai/config_default.toml`. Sane defaults.
- Default field values

Upon further evaluation, we could merge `config_local.toml` and `config_global.toml`. WDYT?

build/config_example.toml
```diff
@@ -51,10 +51,14 @@ solution_detectors = "naive"
...
 
+# Only set this if you want to use a different incident store than the default.
+# If you are running it using podman compose, you should probably leave this
+# alone.
+
 [incident_store.args]
 provider = "postgresql"
 
-host = "127.0.0.1"
+host = "127.1.0.1"
```
jwmatthews, line 61: Is this desired, `127.1.0.1` ?

JonahSussman: Fixed as you wrote it :joy:

build/entrypoint.sh
JonahSussman: I updated how this works to make more sense. See my comment below.

compose.yaml
```diff
@@ -1,47 +1,56 @@
 version: "3"
 
+# TODO: Unify hub importer variables with kai variables. They are currently not
+# stored on KaiConfig.
+
 x-kai-variables: &kai-variables
-  DEMO_MODE: "True"
```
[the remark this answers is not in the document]
JonahSussman: I'm more comfortable with this now that we can also override the config in using user-supplied `config.toml`s. Changing.

build/Containerfile
jwmatthews: Thank you, makes sense.

JonahSussman:
Going to hold off on merging due to #332 preventing me from testing. It was working earlier but after deleting the logs directory, which was having permission errors on bare metal, I can no longer run `podman compose up`. Very very frustrating.

If anyone else can verify everything works then I'm good to merge.

jwmatthews (approved):
Confirmed working on MacOS

jwmatthews:
Closes https://github.com/konveyor/kai/issues/264

JonahSussman:
Confirmed working on Ubuntu 24.04.1

PR #349, opened 2024-09-03
Update demo.md with the correct link to the vscode plugin

cmendible:


jwmatthews (approved):
Thank you @cmendible for the fix!

PR #362, opened 2024-09-13
Repo level awareness 1

JonahSussman:

PR #365, opened 2024-09-16
Runnable codeplan loop

fabianvf:
- Adds Maven validator
- Separates API and codeplan loop
- Plus the validator in so it's ready to feed to an agent

May supersede #362


shawn-hurley (approved):
Looks good to me to get started

Issue #324, opened 2024-08-29
Creation of compiler task runner


dymurray:
```[tasklist]
### Tasks
- [ ] https://github.com/konveyor/kai/issues/337
- [x] Have compiler agent attempt a fix via LLM
- [x] Compiler agent should properly determine whether the code is compilable or not and pass errors to successive agents
```

shawn-hurley:
I believe that this fixed and closing this. 

fixed by #371 and #387 together.

PR #371, opened 2024-09-17
:sparkles: Adding code plan compilation error agent

shawn-hurley:
Adding the compiler agent to the agent's workflow in a notebook.


playpen/repo_level_awareness/agents/compiler_agent.py
```diff
@@ -0,0 +1,138 @@
...
+    # Ouput Instructions 
+    Structure your output in Markdown format such as:
+
+    ## Updated Java File
+    Rewrite the java file here
+
+    ## Reasoning 
+    Write the step by step reasoning in this markdown section. If you are unsure of a step or reasoning, clearly state you are unsure and why. 
+
+    ## Addition Infomation (optional)
```
JonahSussman, line 52: Nit: Should this be "Additional"? Same for the parse_llm_response function below.

playpen/repo_level_awareness/codeplan.py
```diff
@@ -86,6 +86,9 @@ def execute_task(self, task: Task) -> TaskResult:
 
     def get_agent_for_task(self, task: Task) -> Agent:
         for agent in self.agents:
+            print(agent)
+            print(task)
+            print("here!!!!")
```
JonahSussman, line 91: Nit: Unsure if you meant to leave this here

JonahSussman (approved):
Looks good to me, just a few nits here and there

fabianvf (changes requested):
Let's not include coolstore in git, we should use the pattern like in the example directory and clone it/add a script to make changes if necessary. If we want a good example to commit we should make it a minimal POC



Awesome that this is working though

dymurray:
Fixes #324

playpen/repo_level_awareness/agents/compiler_agent.py
shawn-hurley: updated

playpen/repo_level_awareness/codeplan.py
shawn-hurley: updated

fabianvf (approved):

PR #376, opened 2024-09-18
:sparkles: importer: Pass auth token.

jortel:
Update the importer to pass the auth token to the hub.


fabianvf:
@jortel we're using a tool called trunk to enforce formatting, instructions for installation are here. https://docs.trunk.io/code-quality/setup-and-installation/initialize-trunk

PR #377, opened 2024-09-19
:sparkles: Added "VFS" based on git

JonahSussman:
I'd like to use GitPython, but using custom a work-tree and git-dir with is too hard.

As far as git licensing, as long as we don'd modify the source code and provide adequate attribution, I[ don't think there will be a problem](https://opensource.stackexchange.com/questions/7737/am-i-allowed-to-provide-compiled-binaries-of-open-source-libraries) as it git uses the LGPLv2 license.

### DIFF
--- a/playpen/repo_level_awareness/git_vfs.py
+++ b/playpen/repo_level_awareness/git_vfs.py
@@ -168,6 +168,11 @@ class RepoContextManager:
         self.reflection_agent = reflection_agent
 
     def commit(self, msg: str | None = None, spawning_result: Any | None = None):
+        """
+        Commits the current state of the repository and updates the snapshot.
+        Also runs the reflection agent validate the repository state.
+        """
+
         potential_errors = self.reflection_agent.do_whatever_you_need_to_do(
             "argument_that_you_need",
             "another_argument_that_you_need",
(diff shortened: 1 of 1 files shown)


pranavgaikwad (approved):

playpen/repo_level_awareness/git_vfs.py
```diff
@@ -0,0 +1,266 @@
...
+
+        self.reflection_agent = reflection_agent
+
+    def commit(self, msg: str | None = None, spawning_result: Any | None = None):
+        """
+        Commits the current state of the repository and updates the snapshot.
+        Also runs the reflection agent validate the repository state.
+        """
+
+        potential_errors = self.reflection_agent.do_whatever_you_need_to_do(
```
fabianvf, line 176: I'm assuming this is WIP? do we need to establish what this API should look like?

fabianvf (commented):
This is awesome

playpen/repo_level_awareness/git_vfs.py
JonahSussman: I think the semantics of this depends on what the reflection agent needs. I'll defer to Pranav

PR #375, opened 2024-09-18
Add metadata to tracing

devjpt23:
**Capture Number of Tokens in Request and Response**
When this PR is merged, you will be able to capture metadata from the responses into the trace directory when you run [run_demo.py](https://github.com/konveyor/kai/blob/main/example/run_demo.py) . The structure will be as follows:
```bash
── trace
      └── gpt-3.5-turbo << MODEL ID>>
          └── coolstore << APP Name >>
              ├── pom.xml << Source File Path >>
              │   └── single_group << Incident Batch Mode >>
              │       └── 1719673609.8266618 << Start of Request Time Stamp >>
              │           ├── 1 << Incident Batch Number >>
              │           │   ├── 0 << Retry Attempt  >>
              │           │   │   ├── llm_result << Contains the response from the LLM prior to us parsing >>
              │           │   │   ├── token_usage.json<< New metadata file added here >>
              │           │   │   ├── prompt << The formatted prompt prior to sending to LLM >>
              │           │   │   └── prompt_vars.json << The prompt variables which are injected into the prompt template >>
              │           │   ├── params.json << Request parameters >>
              │           │   └── timing << Duration of a Successful Request >>
              └── src
                  └── main
                      ├── java
                      │   └── com
                      │       └── redhat
                      │           └── coolstore
                      │               ├── model
                      │               │   ├── InventoryEntity.java
                      │               │   │   └── single_group
                      │               │   │       └── 1719673609.827135
                      │               │   │           ├── 1
                      │               │   │           │   ├── 0
                      │               │   │           │   │   ├── llm_result
                      │               │   │           │   │   └── token_usage.json << New metadata file added here >>
                      │               │   │           │   ├── prompt
                      │               │   │           │   └── prompt_vars.json
                      │               │   │           ├── params.json
                      │               │   │           └── timing
                      │               │   ├── Order.java
                      │               │   │   └── single_group
                      │               │   │       └── 1719673609.826999
                      │               │   │           ├── 1
                      │               │   │           │   ├── 0
                      │               │   │           │   │   ├── llm_result
                      │               │   │           │   │   └── token_usage.json << New metadata file added here >>
                      │               │   │           │   ├── prompt
                      │               │   │           │   └── prompt_vars.json
                      │               │   │           ├── params.json
                      │               │   │           └── timing

```
This will aid you in debugging by providing detailed metadata and trace information.

[...]

PR #385, opened 2024-09-22
Change from token_usage to all response metadata

jwmatthews:
This is builds on #375.

I saw while testing that not all providers sent back the same keys in 'response_metadata'.
For example Bedrock has the key 'usage' and IBM BAM and Groq have the key 'token_usage'.

To keep things simple, I'm proposing we log all response_metadata to disk and not look for a specific key.

This PR also updates the server to send back the responses metadata in it's result so the client can access it.


fabianvf:
@jwmatthews LGTM, looks like Jason's change I just merged caused a conflict but I'll merge once that's fixed

jwmatthews:
@fabianvf fixed merge conflicts

PR #390, opened 2024-09-25
:sparkles: Featre/analyzer task runner validator

shawn-hurley:
Just to get the initial code up there


shawn-hurley:
This is looking correct, I think it can merge after fixing the conflicts

THings to note: 
* You need to have the kai-analyzer-rpc binary built and pass it in to code plan
* You need to have the rulesets on disk and pass them in to kai.

There seems to be intermintent errors in the pipe between kai-analyzer-prc and the client. I am looking into these things now, but I think having this in to start building off of for the parent child stuff and fixing a single incident may be worth it even with the issues. 

@fabianvf @JonahSussman @pranavgaikwad

pranavgaikwad (approved):
good PR, I like the re-structuring but at the same time, find the `agent` directory a bit awkward maybe something we could fix in the refactor. But overall makes sense to me.

shawn-hurley:
I agree with you, but also it didn't feel like a task runner lol

shawn-hurley:
We need to do a follow-up based on the above comments. 

Task Runners should not have the specifics of agents in their implementation. Instead they should be using one or more "agents". 

This moves us closer to making the "agents" LLM calls more composable. 

I think for us to do this, we need to have a good interface, between taking in strucuted data vs LLM Output.
s

fabianvf (approved):

JonahSussman (approved):
Agreed wrt "we need to have a good interface". LGTM for now

PR #398, opened 2024-09-27
:sparkles: Add rudimentary way to skip failing tasks until other tasks have been completed

fabianvf:


shawn-hurley (approved):
This is awesome, 

I do wonder if there is a way to not check i the results from running a notebook, as we are using it to drive tests?

fabianvf:
@shawn-hurley I cleared the output from the notebook. You can add a git commit hook to do the same by adding

```bash
#!/bin/bash
for f in $(git diff --name-only --cached); do
    if [[ $f == *.ipynb ]]; then
        jupyter nbconvert --clear-output --inplace $f
        git add $f
    fi
done

if git diff --name-only --cached --exit-code
then
    echo "No changes detected after removing notebook output"
    exit 1
fi
```

to `.git/hooks/pre-commit`. This way we'll never commit the output. We could add a test to check this as well if we wanted to.

PR #407, opened 2024-10-02
:sparkles: make analyzer take initialize

djzager:


kai-analyzer/go.sum
```diff
@@ -67,8 +69,6 @@ github.com/pmezard/go-difflib v1.0.0 h1:4DBwDE0NGyQoBHbLQYPwSUPoCMWR5BEzIk/f1lZb
 github.com/pmezard/go-difflib v1.0.0/go.mod h1:iKH77koFhYxTK1pcRnkKkqfTogsbg7gZNVY4sRDYZ/4=
 github.com/sergi/go-diff v1.3.1 h1:xkr+Oxo4BOQKmkn/B9eMK0g5Kg/983T9DqqPHwYqD+8=
 github.com/sergi/go-diff v1.3.1/go.mod h1:aMJSSKb2lpPvRNec0+w3fl7LP9IOFzdc9Pa4NFbPK1I=
-github.com/shawn-hurley/java-external-provider v0.5.4-0.20240904135407-b0958af4ebdf h1:u+zYvsg0ZgpopZtRXltaL+CfTXHVXib+hWhU4F90W/g=
```
shawn-hurley, line 70: big +1

kai-analyzer/pkg/service/analyzer.go
```diff
@@ -22,84 +22,91 @@ type Analyzer struct {
...
+func NewAnalyzer(log logr.Logger) *Analyzer {
+	return &Analyzer{
+		Logger:          log,
+		initedProviders: map[string]provider.InternalProviderClient{},
+	}
+}
+
+func (a *Analyzer) Initialize(args InitializeArgs) error {
 	ctx, cancelFunc := context.WithCancel(context.Background())
 	eng := engine.CreateRuleEngine(ctx,
```
shawn-hurley, line 36: Does this have to be create per initialized thing, or can it be shared?

kai-analyzer/pkg/service/analyzer.go
```diff
@@ -22,84 +22,91 @@ type Analyzer struct {
...
-		"java":    jProvider,
-		"builtin": bProvider,
+	_, err = builtinProvider.ProviderInit(context.Background(), []provider.InitConfig{{Location: args.Location}})
+	if err != nil {
+		cancelFunc()
+		return err
 	}
+	a.initedProviders["builtin"] = builtinProvider
+
+	a.engine = eng
```
shawn-hurley, line 84: Are we ever going to have more than one initial requests?

pranavgaikwad: I imagine each project creating a new instance so no

kai-analyzer/pkg/service/analyzer.go
djzager: I wanted to make the server startup as dumb as possible, accepting all configuration items on the "initialize" request after the server has started. It was my understanding that, when you start gopls you would tell it things like "location" on the `initialize` method call. Maybe I need to research it more thoroughly.

kai-analyzer/pkg/service/analyzer.go
djzager: We should not, although we are not explicitly forbidding it yet. I modeled this exchange after LSP where, if you need to re`initialize` you would have to call `shutdown` and `exit` before calling `initialize` again.

djzager:
I'm dropping this for now. Too massive a change for sprint 1.

PR #408, opened 2024-10-02
Add a hierarchical task structure

### LINKED ISSUE #409: Build priority-aware task manager to focus on relevant work/changes first

fabianvf:
Rather than a simple list of tasks, maintain a priority queue of Task stacks, where new issues spawned from work on a task are added as children of the main task. This allows us to focus on the most relevant issues first, and gives us knobs to play with by being able to tweak the priority of individual tasks/task classes

closes #409

### DIFF
--- a/playpen/repo_level_awareness/api.py
+++ b/playpen/repo_level_awareness/api.py
@@ -1,5 +1,5 @@
 from abc import ABC, abstractmethod
-from dataclasses import dataclass
+from dataclasses import dataclass, field
 from pathlib import Path
 from typing import List, Optional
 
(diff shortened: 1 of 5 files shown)


shawn-hurley (approved):
generally, approve of the PR. Can we move out of draft to merge?

shawn-hurley:
As a side note, is there a way to get the initial_stack, and be able to print out that stack? 

I am thinking about at demo where:
* We get the inital validations and the stack. we can verify this. 
* We fix a particular task, and verify the results.

Does that make sense?

fabianvf:
> As a side note, is there a way to get the initial_stack, and be able to print out that stack?
> 
> I am thinking about at demo where:
> 
>     * We get the inital validations and the stack. we can verify this.
> 
>     * We fix a particular task, and verify the results.
> 
> 
> Does that make sense?

Sure, all of the fields on the TaskManager are accessible, so you can call `get_next_task` outside of a loop. ie:

```python
tm = TaskManager(...)

# Needed so that TM initializes everything
task1 = tm.get_next_task()

highest_priority = min(tm.task_stacks.keys())
initial_task_stack = tm.stack_tasks[highest_priority]
print(initial_task_stack)
```

playpen/repo_level_awareness/api.py
```diff
@@ -14,24 +14,51 @@ class RpcClientConfig:
...
+        )
+
+    def __hash__(self):
+        return hash((self.file, self.line, self.column, self.message))
+
+
+# FIXME: Might not need
+@dataclass
+class TaskResult:
+    encountered_errors: list[str]
```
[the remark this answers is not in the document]
fabianvf: like what?

playpen/repo_level_awareness/codeplan.py
```diff
@@ -112,127 +115,203 @@ def __init__(
...
+
+    def add_task_to_stack(self, task: Task):
+        priority = task.priority
+        if priority not in self.task_stacks:
+            self.task_stacks[priority] = []
+            logger.debug("Created new task stack for priority %s.", priority)
+        self.task_stacks[priority].append(task)
+        logger.debug("Task %s added to priority %s stack.", task, priority)
+
+    def pop_task_from_highest_priority(self) -> Task:
```
JonahSussman, line 220: Small performance thing, but since `Task` has `__lt__`, you should be able to use [Python's built-in priority queue stuff](https://docs.python.org/3/library/heapq.html), if you want.

JonahSussman (approved):
Makes sense, only one small thing that you might want to use [heapq](https://docs.python.org/3/library/heapq.html) for performance, but that's non-essential for now.

playpen/repo_level_awareness/codeplan.py
fabianvf: Yeah I considered it but thought it was pretty easy to manage myself, but it still could be worth moving to the implementation for terseness/performance. Hopefully the queue doesn't grow that large...

PR #410, opened 2024-10-03
:sparkles: Adding dependency agent

shawn-hurley:
Notes: That will make the LLM respond with function calls rather than rewrite the full file.

The dependency agent defines three functions for the LLM that it can use to write the code to "fix" the dependencies. 

* search_fqdn is responsible for searching the dependency in Maven Central. Due to the restricted environment, we may eventually need to make something that can also search for looking up t.
* find_in_pom is responsible for finding an existing dependency in the pom.xml. We do need to figure out how to handle sub-modules in the future. For now, we will just work with the parent pom.
* editor is responsible for actually making the changes. This is actually the responsibility of the task runner, and we may not actually use this code, as we may just be able to use the output of the first two calls. 

There is a lot of parsing that has to be done with this approach, and I think that we may be able to write something that is way more generic in the future if this is approach is something that we want to go with. 

**Note: that this is still a work in progress and the tests are not all passing at the time of posting**


playpen/repo_level_awareness/agent/dependency_agent/api.py
```diff
@@ -0,0 +1,41 @@
...
+
+
+@dataclass
+class FQDNResponse:
+    artifact_id: str
+    group_id: str
+    version: str
+
+    def to_llm_message(self) -> HumanMessage:
+        return HumanMessage(f"the result is {json.dumps(self.__dict__)}")
```
fabianvf, line 17: Does the `__dict__` have any cruft in it or just what you want?

playpen/repo_level_awareness/agent/dependency_agent/dependency_agent.py
```diff
@@ -0,0 +1,347 @@
+# trunk-ignore-begin(ruff/E402)
+import sys
+
+sys.modules["_elementtree"] = None
```
fabianvf, line 4: Mind leaving a comment for why this is necessary?

playpen/repo_level_awareness/agent/dependency_agent/dependency_agent.py
```diff
@@ -0,0 +1,347 @@
...
+            # Break out of the while loop, if we don't have a final answer then we need to retry
+            if not self.__should_continue(llm_response):
+                break
+
+            # We do not believe that we should not conintue now we have to continue after running the code that is asked to be run.
+            # The only exception to this rule, is when we actually update the file, that should be handled by the caller.
+            # This happens sometimes that the LLM will stop and wait for more information.
+
+            for a in llm_response.actions:
+                for s, method in self.agent_methods:
```
fabianvf, line 213: nit but maybe use `method_name` instead of `s` just for readability

playpen/repo_level_awareness/agent/dependency_agent/dependency_agent.py
```diff
@@ -0,0 +1,347 @@
...
+            llm_response = self.parse_llm_response(fix_gen_response.content)
+            # Break out of the while loop, if we don't have a final answer then we need to retry
+            if not self.__should_continue(llm_response):
+                break
+
+            # We do not believe that we should not conintue now we have to continue after running the code that is asked to be run.
+            # The only exception to this rule, is when we actually update the file, that should be handled by the caller.
+            # This happens sometimes that the LLM will stop and wait for more information.
+
+            for a in llm_response.actions:
```
fabianvf, line 212: nit but maybe use `action` instead of `a`  just for readability

playpen/repo_level_awareness/agent/dependency_agent/dependency_agent.py
```diff
@@ -0,0 +1,347 @@
...
+                    case "Final Answer":
+                        actions.append(
+                            _action(code_block, thought_str, observation_str)
+                        )
+                        in_final_answer = True
+                        in_code = False
+                        in_thought = False
+                        continue
+
+            # TODO: There has to be a better way with python to do this.
```
fabianvf, line 321: something like

```python
# Initialize the sections with their respective flags and configurations
sections = {
    'code_block': {'lines': [], 'strip_line': True, 'active_flag': in_code},
    'final_answer': {'lines': [], 'strip_line': False, 'active_flag': in_final_answer},
    'thought_str': {'lines': [], 'strip_line': True, 'active_flag': in_thought},
    'observation_str': {'lines': [], 'strip_line': False, 'active_flag': in_observation},
}

# Process each line and append to the appropriate section
for line in lines:
    for section in sections.values():
        if section['active_flag']:
            line_content = line.strip() if section['strip_line'] else line
            section['lines'].append(line_content)

# Construct the final content for each section
code_block = '\n'.join(sections['code_block']['lines']).strip()
final_answer = '\n'.join(sections['final_answer']['lines']).strip()
thought_str = '\n'.join(sections['thought_str']['lines']).strip()
observation_str = '\n'.join(sections['observation_str']['lines']).strip()
```

could work and do the same?

playpen/repo_level_awareness/agent/dependency_agent/dependency_fqdn_selection.py
```diff
@@ -0,0 +1,142 @@
...
+Do not use a dependency that has already been tried.
+                       
+Think through the problem fully. Do not update the dependency if it has moved to newer versions; we want to find the version that matches, regardless of whether it is old or not. 
+
+Output in the format of:
+
+ArtifactId:
+GroupId:
+
+Reasoning
```
fabianvf, line 58: I'd put reasoning first

playpen/repo_level_awareness/task_runner/dependency/task_runner.py
```diff
@@ -0,0 +1,111 @@
...
+        DependencyValidationError,
+        SymbolNotFoundError,
+        PackageDoesNotExistError,
+    )
+
+    def __init__(self, agent: MavenDependencyAgent) -> None:
+        self._agent = agent
+
+    def can_handle_task(self, task: Task) -> bool:
+        return isinstance(task, self.handeled_type)
```
fabianvf, line 53: `handeled` -> `handled`

fabianvf (commented):
Nothing major but there are some single letter variables that make it a little hard to read and some odd things that could use a comment explaining

(I didn't bother annotating every single letter variable but I can if you want)

playpen/repo_level_awareness/agent/dependency_agent/dependency_agent.py
fabianvf: obviously you'd initialize outside the loop and just set the flags manually

fabianvf: I don't think this is necessarily better though

playpen/repo_level_awareness/agent/dependency_agent/api.py
shawn-hurley: IIUC `self.__dict__` just makes the data class a dict, so yes everything in the dataclass is what I want

If I am wrong please let me know, this was a lot of google -> code as I try and re-learn python

playpen/repo_level_awareness/agent/dependency_agent/dependency_agent.py
shawn-hurley: +1

pranavgaikwad (approved):

playpen/repo_level_awareness/agent/dependency_agent/dependency_agent.py
shawn-hurley: sorry forgot we were in python I think for a little bit

playpen/repo_level_awareness/agent/dependency_agent/dependency_agent.py
shawn-hurley: I think that we are going to leave it, and come back to see if that makes sense?

Going to resolve go ahead and re-open it if you disagree

JonahSussman (approved):
LGTM. I _think_ this changes some things with the RPC server but that's ok

playpen/repo_level_awareness/agent/dependency_agent/api.py
fabianvf: looks like that's accurate if you only have strings/no nested values. There's also `dataclass.asdict` which handles edge cases but this is probably fine

PR #413, opened 2024-10-04
Change config precedence so env vars are king

### LINKED ISSUE #399: Precendence of environment variables or configuration file values is causing unexpected issues
As I am testing expected usage patterns of `podman compose up` I've run into 2 confusion points when a user supplies a `config.toml` to run with.

- #395 
- #397 

The issue in both of those cases is that for `podman compose up` we are specifying values in https://github.com/konveyor/kai/blob/main/compose.yaml which are intended to take highest precedence.  These values are transmitted via environment variables to the Kai process.

When we run with a specified configuration file, the path to the configuration file is being passed in as a command line argument in the entrypoint of the container:
https://github.com/konveyor/kai/blob/main/build/entrypoint.sh#L45

	if [[ -f /podman_compose/build/config.toml ]]; then
		printf "Using custom config file\n"
		PYTHONPATH="/kai/kai" python /kai/kai/server.py --config_filepath /podman_compose/build/config.toml
	else
		PYTHONPATH="/kai/kai" python /kai/kai/server.py
	fi

As this configuration file appears to be an explicit command line argument via the entrypoint we are interpreting the values in the configuration file to take higher precedence than what we have in environment variables.   

In the case of `podman compose up` this feels wrong, I think he User will assume that compose.yaml is the first source of truth.
I think they will assume the precedence is:
1. Environment variables
2. Configuration file

I don't think they will think that the configuration file they use ends up being a command line argument so it takes higher precedence. 

When I think of the user running `podman compose up`, my assumptions of their desires are: 
- `compose.yaml` takes precedence, this is where we set things like DB host, Demo Mode, Logging, etc
- `build/config.toml` is used to change models and/or perhaps algorithm choices, but primarily I see model choices as the main setting being changed

dymurray:
fixes #399

### DIFF
--- a/kai/models/kai_config.py
+++ b/kai/models/kai_config.py
@@ -247,15 +247,15 @@ class KaiConfig(BaseSettings):
         Config is loaded with the following priority (higher overrides lower):
 
         - Command line args (not implemented)
-        - Config file that is declared on the command line / via init arguments.
         - Environment vars
+        - Config file that is declared on the command line / via init arguments.
         - Global config file (kai.config.toml)
         - Default field values
         """
         return (
-            init_settings,
             env_settings,
             dotenv_settings,
+            init_settings,
             file_secret_settings,
             TomlConfigSettingsSource(
                 settings_cls, os.path.join(PATH_KAI, "config.toml")


dymurray:
After talking to @JonahSussman We may prefer `config.toml` taking precedence. If we do we just need to remove some of the defaults in `config_example.toml` to resolve the podman issues.

dymurray:
@shawn-hurley personally I would like to see this merged. I don't think that changing the example file is enough to resolve this the more I play with it. If I switch between dev env and the compose workflow this will still break with the database config values set.

I feel strongly that env vars taking highest precedence is the most common flow of configuration across languages. Am I wrong?

JonahSussman:
I'm down to make env variables king, but the PR currently works by overriding the init function of the class. Take, for example, this environment:

```
export KAI__LOG_LEVEL = INFO
```

and you did this in Python in a Python test case:

```
config = KaiConfig(log_level="DEBUG")
assert config.log_level == "DEBUG"
```

It will fail in some environments and succeed in others. Thus, I don't think we should mess with the precedence. Instead, I think we should:

1. Remove the build/config.toml file entirely
2. Ask users to configure their instance with a .env file, and pass it in to podman compose with --env-file <file>  (pydantic-settings should support natively, but podman just might make it part of the environment so it doesn't matter. Haven't tested this)
3. (Optional) Somehow expose command line arguments to the user outside of podman compose up so the user can override things to their heart's content

Thoughts?

JonahSussman:
To clarify, the crux of the issue is that you need to have some way to pass in config through `podman compose` to the main program. Whether that's a `build/config.toml`, environment variables, or a `.env` file, there's no way to avoid it.

We would basically be swapping out the `build/config.toml` approach with a `.env` approach. So the config file:

```toml
[models]
provider = "ChatIBMGenAI"

[models.args]
model_id = "mistralai/mixtral-8x7b-instruct-v01"
```

would become:

```
KAI__MODELS__PROVIDER = "ChatIBMGenAI"
KAI__MODELS__ARGS__MODEL_ID = "mistralai/mixtral-8x7b-instruct-v01"
```

shawn-hurley:
I don't know if I would want to remove a config file. Env vars in the Windows world are a lot harder to work, IIRC. 

In Podman compose, can you not mount the configuration file?

dymurray:
Closing in favor of a different approach

PR #416, opened 2024-10-09
:seedling: just parameterize the hardcoded things

djzager:
I am limiting my changes as much as possible, just to get this to work from the VSCode extension.


kai-analyzer/main.go
```diff
@@ -29,18 +31,22 @@ func main() {
 		panic(fmt.Errorf("rules directory must be valid"))
 	}
 
-	file, err := os.Create("kai-analyzer.log")
-	if err != nil {
-		panic(err)
-	}
-	logger := slog.NewJSONHandler(file, &slog.HandlerOptions{
+	logger := slog.NewJSONHandler(os.Stderr, &slog.HandlerOptions{
```
shawn-hurley, line 34: why not a file?

djzager: because I can just stream stderr directly to the VSCode outputchannel

shawn-hurley: Lots of other LSP's end up using a file, and because we will not just be using it from the LSP, but from KAI as well, what would be the donwside of using a file?

djzager: I think, in the end that is fine, right now streaming it directly to the console of vscode has been incredibly convenient.

pranavgaikwad: does a combined handler work?

PR #417, opened 2024-10-09
:sparkles: Allows issues to be seeded at max priority

fabianvf:
- Add a `seed_tasks` argument to TaskManager which are added at priority 0
- Add a `max_priority` argument to TaskManager.get_next_task, which ends the iteration once no more issues below that priority are found
- Checks if tasks are incidentally solved and removes them from the queue if they are
- Removes unused `updated_file_content` argument from TaskManager
- Adds `creation_order` to Task objects, purely for my own sanity in deterministic testing


shawn-hurley (approved):
This looks really good to me!

PR #422, opened 2024-10-14
:seedling: kai-analyzer: use ServeCodec

djzager:


kai-analyzer/main.go
```diff
@@ -54,7 +55,20 @@ func main() {
...
+	// Check if Java exists on the PATH
+	if err := exec.Command("java", "-version").Run(); err != nil {
+		panic("Java is not installed or not on the PATH")
+	}
+	l.Info("Java is installed")
+
+	// Check if Maven exists on the PATH
+	if err := exec.Command("mvn", "-version").Run(); err != nil {
+		panic("Maven is not installed or not on the PATH")
+	}
```
djzager, line 68: yay or nay on this? This, while not the ultimate problem with the extension integration, would have been really helpful early on when I hadn't installed these.

shawn-hurley: I like it, this is a good idea.

We even may want to consider logging the versions because they might have some known issues if you are using older Java or Maven.

shawn-hurley (approved):

PR #430, opened 2024-10-16
:ghost: Move out of playpen and add mypy type checking Github action

JonahSussman:
[Trunk check also supports mypy](https://docs.trunk.io/code-quality/linters/supported/mypy) if we want to go that route. I'd also like to move the backend server into its own folder but that can be another discussion.


shawn-hurley (approved):
just would like to see the type check pass as well as wondering why a different script for mypy instead of adding to trunk

pranavgaikwad (approved):

PR #431, opened 2024-10-16
:ghost: Fixed some stuff

JonahSussman:
Summary of fixes:
- Made `dangerously_reset_coolstore.sh` work properly
- Made jsonrpc library support arrays as params
- Reverted ast_diff __iter__ method
- Added ValidationException
- Changed playpen agent notebook to use tmp dir inside its directory as my OS was clobbering the /tmp/ directory it was making


shawn-hurley (approved):
let's make sure to remove the output from the notebook before we merge; otherwise, it looks good to me. 

Do we want to re-enable the trace logs, so that we can figure that out later?

JonahSussman:
@shawn-hurley Sure I'm down to re-enable them. Clearing the output and then will merge.

Issue #350, opened 2024-09-04
Consistent: "No codeblocks detected in LLM response" for several files with 


jwmatthews:
I am seeing consistent and repeatable issues with several files in Coolstore when I run against claude 3.5 sonnet.
It looks like the output stops suddenly midway through generating an update.

Config:
```
[models]
provider = "ChatBedrock"

[models.args]
model_id = "anthropic.claude-3-5-sonnet-20240620-v1:0"

```

Error snippet:
```
WARNING - 2024-09-04 06:50:33,902 - kai.models.file_solution - [    file_solution.py:95   - parse_file_solution_content()] - No codeblocks detected in LLM response
WARNING - 2024-09-04 06:50:33,907 - kai.service.kai_application.kai_application - [  kai_application.py:202  - get_incident_solutions_for_file()] - Request to model failed for batch 1/1 for src/main/java/com/redhat/coolstore/model/ShoppingCart.java with exception, retrying in 10.0s
Error in LLM Response: The LLM did not provide an updated file for src/main/java/com/redhat/coolstore/model/ShoppingCart.java

```

Attempting to convert:

- https://github.com/konveyor-ecosystem/coolstore/blob/main/src/main/java/com/redhat/coolstore/service/ShoppingCartService.java

prompt:

- https://gist.github.com/jwmatthews/c1dc0e251a9bd40e1efb83172e383439

llm_result (all failures, stops prematurely)

- https://gist.github.com/jwmatthews/7b444fadd6db8645f1803f7bb16c8702
- https://gist.github.com/jwmatthews/181e4c1ba3ccb3c98159f51c9ff5f58b
- https://gist.github.com/jwmatthews/7b21d4b6135e569d490c91de66266144
- https://gist.github.com/jwmatthews/554dcc24809141bd549f2c5e710995fe
- https://gist.github.com/jwmatthews/9dd55a9378ed7437074d97fae0132f25

Note on a subsequent retry it failed once more and then succeeded but the contents of what it generated are incomplete/truncated

1 more failure: https://gist.github.com/jwmatthews/7d7aac70a6b69291e2ff0ed2b467debb

Partial Success but Incomplete:  https://gist.github.com/jwmatthews/0b366ffa4ff8fe2ed89638552e9972e9
It truncates the response and adds a comment  ` // Rest of the class remains unchanged`

```java
package com.redhat.coolstore.service;

import java.util.Hashtable;
import java.util.logging.Logger;

import jakarta.ejb.Stateful;
[...]

PR #436, opened 2024-10-20
Fix no codeblock error by re-prompting to LLM's incomplete response

devjpt23:
The "No codeblocks" as faced in earlier in [issue 350](https://github.com/konveyor/kai/issues/350). The issue arises when the LLM responds without including two sets of triple backticks (```) in the updated file content.

To address this, this PR modifies`kai_application.py` so that if a response lacks proper code blocks and triggers the message "No codeblocks...", the system will automatically re-invoke the LLM, requesting a complete and properly formatted response.

### DIFF
--- a/kai/kai_trace.py
+++ b/kai/kai_trace.py
@@ -129,24 +129,35 @@ class KaiTrace:
 
     @enabled_check
     def llm_result(
-        self, current_batch_count: int, retry_count: int, result: BaseMessage
+        self,
+        current_batch_count: int,
+        retry_count: int,
+        result: BaseMessage,
+        output_filename: str,
     ):
         result_file_path = os.path.join(
-            self.trace_dir, f"{current_batch_count}", f"{retry_count}", "llm_result"
+            self.trace_dir,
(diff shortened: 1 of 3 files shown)


jwmatthews:
@devjpt23 thank you for working on this issue.  Due to unfortunate timing as you were working this we moved to agentic workflows that resulted in an almost rewrite of all of Kai.  

I think the best path would be to close this PR #436 and PR #458.

If you were interested to continue to work on this problem area, I recommend getting familiar with how Kai is being run from the IDE via https://github.com/konveyor/editor-extensions.  Then as you see the workflow, you could work with team to either reopen these PRs and update them to the newer codebase or maybe just use the knowledge you've gained to find the right solution with the new approach.  

For now I will close #436 and #458 

Thank you for the time spent, sorry we weren't able to merge these 2 PRs.

devjpt23:
@jwmatthews Thank you for the opportunity. I am very interested in working on the new approach using agents. However, I am still figuring out how to run it via the IDE. Once I have that sorted, I will start looking into the agents more carefully.

PR #438, opened 2024-10-21
:sparkles: Adding Logger that enforces log level

shawn-hurley:
Disallow for an individual component to change the level of the log for the program; instead, force the level set in the configuration file to all the logs.

I will complete the tracing after this merges and after #451 is merged.


kai/logging/kai_trace.py
JonahSussman, line 1: Not a huge fan of how KaiTrace was implemented in the past. We should probably think of a better way to do tracing later.

kai/logging/logging.py
```diff
@@ -0,0 +1,138 @@
...
+
+
+class KaiLogger(logging.Logger):
+    configLogLevel: int
+
+    def __init__(self, name: str, configLogLevel: int) -> None:
+        super().__init__(name, configLogLevel)
+        self.configLogLevel = configLogLevel
+
+    @override
```
JonahSussman, line 17: Not sure if `@override` is necessary

playpen/old_rpc_stuff/client/rpc.py
JonahSussman, line 1: I think we can probably remove `old_rpc_stuff`, no?

tests/test_kai_logging.py
```diff
@@ -77,30 +77,9 @@ def test_init_logging(self):
...
-        mock_getcwd.return_value = base_dir
-        expected_log_dir = os.path.join(base_dir, "logs")
-        init_logging_from_config(self.config)
-
-        mock_makedirs.assert_called_once_with(expected_log_dir, exist_ok=True)
-
-        logger = logging.getLogger("kai")
+        logger = get_logger("test")
         self.assertEqual(logger.level, logging.DEBUG)
+        print(f"herelrhelajdlfkjadlkjfsdlfjkasdlfkj {logger.handlers}")
```
JonahSussman, line 82: lol

JonahSussman (approved):
A few minor things

pranavgaikwad (approved):

PR #442, opened 2024-10-23
Building off of Fabian and Jonah's work

shawn-hurley:
I have fixed the types and the tests, as well as the linter issues. You should be good to!


fabianvf (approved):
Nice!

PR #443, opened 2024-10-23
Project Restructure

shawn-hurley:
This is our attempt at re-structuring the project. 

There are some TODO's left around:
* config
* Tracing
* Logging
* Some general more clean up
* We should also have a conversation around the explicit exporting of things. I removed some of this to get the type system happy, as there was some errant warnings on "re-exporting"
* There is also some shared python library stuff between solution server and the kai binary, we may want to consider a top level directory, or make the kai-rpc server it's own top level, and use `kai` as a library for all the actual top level things (kai-rpc-server, kai-solution-server).
* We also want to refactor to remove the kai stutter from the imports


kai/reactive_codeplanner/task_runner/compiler/compiler_task_runner.py
```diff
@@ -146,6 +146,11 @@ def execute_task(self, rcm: RepoContextManager, task: Task) -> TaskResult:
     def parse_llm_response(self, message: BaseMessage) -> MavenCompilerLLMResponse:
         """Private method that will be used to parse the contents and get the results"""
 
+        if isinstance(message.content, list):
+            return MavenCompilerLLMResponse(
+                reasoning="", java_file="", additional_information=""
```
fabianvf, line 151: don't quite follow this

kai/reactive_codeplanner/task_runner/dependency/task_runner.py
```diff
@@ -67,6 +68,9 @@ def execute_task(self, rcm: RepoContextManager, task: Task) -> TaskResult:
         maven_dep_response = self._agent.execute(MavenDependencyRequest(task.file, msg))
         logger.info("got mvn dep response: %r", maven_dep_response)
 
+        if not isinstance(maven_dep_response, MavenDependencyResult):
```
fabianvf, line 71: I feel like we should probably error here or something right?

kai/reactive_codeplanner/task_runner/dependency/task_runner.py
```diff
@@ -91,15 +95,18 @@ def execute_task(self, rcm: RepoContextManager, task: Task) -> TaskResult:
         root = tree.getroot()
         deps = root.find("{http://maven.apache.org/POM/4.0.0}dependencies")
 
+        if deps is None or not isinstance(deps, ET.Element):
```
fabianvf, line 98: I'm just worried the empty returns when we see something wrong is going to be very confusing. We should probably be raising on these and add some error handling to the TaskManager

kai_solution_server/service/incident_store/incident_store.py
```diff
@@ -447,7 +454,9 @@ def post_process(self, limit: int = 5) -> None:
         page_size = 100
         page = 0
         processed_count = 0
-        limit = float("inf") if limit < 0 else limit
+        # I have no idea what this is trying to do.
+        # we need a better way.
```
fabianvf, line 458: It's making it so that processed_count will never go above limit if you set limit to a negative number, ie -1 -> infinity. I think. Never seen it before tbh

kai/reactive_codeplanner/task_runner/compiler/compiler_task_runner.py
shawn-hurley: basically, we don't know how to parse llm responses that are not just a string.

we need to handle this in the future

kai_solution_server/service/incident_store/incident_store.py
shawn-hurley: yeah, i think, but the types are different and that messes things up.

kai/reactive_codeplanner/task_runner/dependency/task_runner.py
shawn-hurley: I agree, I think we need to add a task/discussion on how to implement error handeling

kai/reactive_codeplanner/task_runner/dependency/task_runner.py
shawn-hurley: See previous comment, I agree, but want a wholistic approach

kai/reactive_codeplanner/agent/ast_diff/parser.py
```diff
@@ -10,7 +10,7 @@ class Language(Enum):
     Java = "java"
 
 
-def extract_ast_info(root: ts.Node, language: Language) -> DiffableSummary:
+def extract_ast_info(root: ts.Tree, language: Language) -> DiffableSummary:
```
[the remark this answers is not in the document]
shawn-hurley: It does, I verified that Tree is what it really is working on, not Node

kai_solution_server/service/incident_store/incident_store.py
JonahSussman: See my latest commit. It would require changing the while loop but that requires brain power right now (and potentially modifying the functionality)

kai/reactive_codeplanner/agent/ast_diff/parser.py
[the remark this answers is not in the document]
pranavgaikwad: yes

PR #451, opened 2024-10-24
:ghost: Changed agents to use ModelProvider

JonahSussman:


shawn-hurley (changes requested):
If we do this we should move this into Kai/. Or keep in solution service. Especially if we move to rpc server is top level and they both use Kai as a library

shawn-hurley (approved):

kai/llm_interfacing/model_provider.py
```diff
@@ -13,6 +15,10 @@
 
 from kai.kai_config import KaiConfigModels
 
+if TYPE_CHECKING:
```
shawn-hurley, line 18: Why this though?

kai/llm_interfacing/model_provider.py
```diff
@@ -168,6 +174,16 @@ def __init__(self, config: KaiConfigModels):
...
             self.llama_header = config.llama_header
 
+    def invoke(
+        self,
+        input: LanguageModelInput,
+        config: Optional[RunnableConfig] = None,
+        *,
+        stop: Optional[list[str]] = None,
+        **kwargs: Any,
+    ) -> Any:
```
shawn-hurley, line 184: Isn't the output have to be a BaseMessage?

kai/llm_interfacing/model_provider.py
JonahSussman: `TYPE_CHECKING` will only import the types during type checking to avoid circular imports. _Technically_ we should import types that we only use for type checking in one of these blocks.

kai/llm_interfacing/model_provider.py
JonahSussman: You are correct! :astonished:

PR #453, opened 2024-10-24
:sparkles: Restore run_demo

pranavgaikwad:
This PR makes it so that we can run code plan loop without child issues for all incidents in coolstore.

We still have outstanding items: 
1. Before running run_demo, users are supposed to install some dependencies, once @jmontleon is done with his gh action, I will follow up with a script that downloads them. 
2. We have to record all VCR responses.


example/run_demo.py
```diff
@@ -1,253 +1,184 @@
...
+    rpc_server.start()
 
-    # We expect that we are overwriting the file, so all directories should exist
-    intended_file_path = f"{SAMPLE_APP_DIR}/{file_path}"
-    if not os.path.exists(intended_file_path):
-        KAI_LOG.warning(
-            f"**WARNING* File {intended_file_path} does not exist.  Proceeding, but suspect this is a new file or there is a problem with the filepath"
-        )
+    # wait for server to start up
+    time.sleep(3)
```
shawn-hurley, line 93: is there a better way to check this in the future?

shawn-hurley (approved):
This looks good to me I would just as soon merge this in to the feature branch so that we can all start testing on it

JonahSussman (approved):
ACK

PR #458, opened 2024-10-27
Merge prompt code with llm response

devjpt23:
**Problem:** Usually, the LLM provides a complete response, but sometimes it returns an incomplete response, often due to an unnecessary comment like "// ... Rest of the code remains unchanged." This "lazy effect" occurs when the LLM avoids generating code that's identical to the input, assuming it can be inferred. This was experienced in #350.

**Solution:** One option is to re invoke the LLM, prompting it to provide the complete code. However, this approach increases token usage, adding unnecessary cost in the long run. In this PR, we avoid additional invocations by directly merging the code from the prompt with the LLM response wherever the code remains unchanged, achieving the desired result without extra token consumption.

**NOTE:** This PR is dependent on #436.

### DIFF
--- a/kai/kai_trace.py
+++ b/kai/kai_trace.py
@@ -128,7 +128,7 @@ class KaiTrace:
             f.write(json.dumps(data, indent=4))
 
     @enabled_check
-    def llm_result(
+    def llm_result_with_codeblocks(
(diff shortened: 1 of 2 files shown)


jwmatthews:
Closing as per comment in https://github.com/konveyor/kai/pull/436#issuecomment-2622826970

Issue #456, opened 2024-10-25
Add VCR responses to new run_demo that uses codeplan solution


pranavgaikwad:
We need to record responses for all incidents for the codeplan flow

PR #462, opened 2024-10-29
:sparkles: Add VCR to agents and add generated VCR responses for gpt-4o & llama-70b

pranavgaikwad:
This partially addresses #456 

I don't like how I have to pass demo_mode on per Task as each agent is using VCR separately. I think we can bring that all to task_manager but it doesn't have model_provider so it's fine for now (I think). Ideally we should have a generic caching mechanism for everything (not just HTTP responses but also validators).


kai/kai_vcr_util.py
```diff
@@ -0,0 +1,63 @@
...
+
+    record_mode = "once" if demo_mode else "all"
+
+    my_vcr = vcr.VCR(
+        cassette_library_dir=os.path.join(
+            PATH_DATA, "vcr", application_name, agent, model_id
+        ),
+        record_mode=record_mode,
+        match_on=[
+            "uri",
```
fabianvf, line 30: Are these matchers working for multiple calls for the same file?

pranavgaikwad: tbh, I didn't touch the matchers, my assumption was that this was matching based on what we send in http request

fabianvf: The matchers were configured for the old use case, if the request body/content changes it will still trigger a match. If you're reading/writing from the same file and only the request body is changing it won't register it as a different request

pranavgaikwad: ahh okay, thanks for pointing that out

pranavgaikwad: @fabianvf added body to requests for now, seems to be working as expected

kai/reactive_codeplanner/task_runner/analyzer_lsp/task_runner.py
```diff
@@ -136,9 +137,16 @@ def execute_task(self, rcm: RepoContextManager, task: Task) -> TaskResult:
             incidents=[task.incident],
         )
 
-        aimessage = self._model_provider.invoke(
-            [self.system_message, HumanMessage(content=content)]
-        )
+        with playback_if_demo_mode(
```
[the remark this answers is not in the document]
pranavgaikwad: @fabianvf I thought about that too. But task manager doesn't directly take model_provider and we need that for vcr

pranavgaikwad:
We are closing this in favor of a cache at model provider level.

PR #467, opened 2024-11-04

shawn-hurley:

* Remove xml line number parser and moved to lxml for xml
* Changed the way dep matches and overriding of deps was handled To not rely on line numbers
* Changed AgentResult to have defaults and updated a field from a list to a single value. This is because we should only work on a single file at a time, and we don't need more than one file in the agent.
* Added a test case for the AnalyzerAgent


shawn-hurley:
@JonahSussman in the last comit i addressed the feedback on the pr review in #466 let me know how it looks.

PR #470, opened 2024-11-04

shawn-hurley:

Remove xml line number parser and moved to lxml for xml
Changed the way dep matches and overriding of deps was handled To not rely on line numbers
Changed AgentResult to have defaults and updated a field from a list to a single value. This is because we should only work on a single file at a time, and we don't need more than one file in the agent.
Added a test case for the AnalyzerAgent


JonahSussman (approved):
Looks good. I saw you addressed the previous feedback!

kai/reactive_codeplanner/agent/analyzer_fix/__init__.py
```diff
@@ -0,0 +1 @@
+__all__ = ["AnalyzerFixRequest", "AnalyzerFixResponse", "AnalyzerAgent"]
```
fabianvf, line 1: You'll need to import these or else `from kai.reactive_codeplanner.agent.analyzer_fix import *` will just fail. `__all__` lists the symbols in the file which should be exported when `import *` is used, so you just need to import them here. Or don't add it at all, but this way is broken

kai/reactive_codeplanner/agent/dependency_agent/api.py
```diff
@@ -38,10 +35,23 @@ def to_xml_element(self) -> ET.Element:
...
-    end_line: int
+    override: bool
+    group_id: str | None = None
+    artifact_id: str | None = None
+    version: str | None = None
 
     def to_llm_message(self) -> HumanMessage:
         return HumanMessage(
-            f"the start_line is {self.start_line} and end_line is {self.end_line}"
+            ## TODO: I don't think that this will work.
```
fabianvf, line 45: What is this for?

kai/reactive_codeplanner/agent/analyzer_fix/agent.py
```diff
@@ -0,0 +1,195 @@
...
+        lines_of_output = cast(str, message.content).splitlines()
+
+        in_java_file = False
+        in_reasoning = False
+        in_additional_details = False
+        java_file = ""
+        reasoning = ""
+        additional_details = ""
+        for line in lines_of_output:
+            if line.strip() == "## Reasoning":
```
pranavgaikwad, line 152: I think we need to make this more error-proof in future. Strict matching on section headers sometimes fails when LLMs don't quite follow the exact syntax

kai/reactive_codeplanner/agent/dependency_agent/api.py
```diff
@@ -38,10 +35,23 @@ def to_xml_element(self) -> ET.Element:
...
+            "the start_line is 1 and end_line is 2"
         )
+
+    def match_dep(self, dep: ET._Element) -> bool:
+        found = []
+        for child in dep:
+            if child.text in [self.group_id, self.artifact_id, self.version]:
+                found.append(True)
```
fabianvf, line 53: There's something about this I hate but it's fine :joy:

fabianvf (commented):
A few minor issues but mostly lgtm

pranavgaikwad (approved):

kai/reactive_codeplanner/agent/analyzer_fix/__init__.py
shawn-hurley: I think did this to "fix" a mypy re-export issue but I see what you mean. I wonder if there is another thing that I can add to just say "explicitly export these".

kai/reactive_codeplanner/agent/dependency_agent/api.py
shawn-hurley: it is very golang i think, but I also didn't know a better way off the top of my head, if there are suggestions on how to make it more pythonic that would work for me.

kai/reactive_codeplanner/agent/analyzer_fix/agent.py
shawn-hurley: Yeah, I agree. would a followup issue on a more comprehensive LLM response parsing be a good idea?

kai/reactive_codeplanner/agent/analyzer_fix/__init__.py
shawn-hurley: fixed

kai/reactive_codeplanner/agent/analyzer_fix/agent.py
pranavgaikwad: sounds good to me

fabianvf (approved):

PR #471, opened 2024-11-05
:sparkles: Added preliminary cli script

JonahSussman:


JonahSussman:
@shawn-hurley Ah I see. So this is less having something for testing stuff and more having an IDE that an actual end user could use, if I'm understanding correctly?

shawn-hurley:
I don't know about an IDE more like a CLI tool to just "do it" if that makes sense?

PR #476, opened 2024-11-07
:bug: Priority Queue now always returns highest depth tasks for a given priority level first

fabianvf:
- Changes `__lt__` behavior of tasks to do depth then priority
- Fixes the priority queue sort when a task is appended so that it can pop the next task properly
- Adds a `__str__` function to the priority queue for easier introspection during debugging
- Adds timing information and queue state logging to the `reactive_codeplanner/main.py`, to aid debugging

### DIFF
--- a/kai/reactive_codeplanner/main.py
+++ b/kai/reactive_codeplanner/main.py
@@ -1,15 +1,18 @@
 import argparse
-import logging
+import time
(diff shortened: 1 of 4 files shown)


shawn-hurley (approved):
LGTM,

PR #487, opened 2024-11-14
Adding ability for dependency rules to be set correctly in the seed

shawn-hurley:
of the server


fabianvf (approved):
We probably need something a little more robust in the long-term but this looks good for now

PR #488, opened 2024-11-14
:ghost: Add slides from kai-community call

fabianvf:


shawn-hurley:
Can you add a readme, that will show someone how to run the slides as a slideshow and the tool you used?

PR #490, opened 2024-11-14
:bug: Fix dependency agent back and forth chat for not-so-good models

pranavgaikwad:
I think we can simplify the agent a lot because effectively only one of the tools is actually being used but we describe 4 tools in the prompt. But I didn't want to introduce new issues because of it, so I want to take the time to redo / simplify the agent.


fabianvf (approved):
I think this makes sense, @shawn-hurley is more familiar with this agent though.

LGTM

kai/reactive_codeplanner/task_runner/dependency/task_runner.py
```diff
@@ -88,7 +88,7 @@ def execute_task(self, rcm: RepoContextManager, task: Task) -> TaskResult:
         )
         pom = os.path.join(os.path.join(rcm.project_root, "pom.xml"))
         # Needed to remove ns0:
-        ET.register_namespace("", "http://maven.apache.org/POM/4.0.0")
+        # ET.register_namespace("", "http://maven.apache.org/POM/4.0.0")
```
shawn-hurley, line 91: Lets just remove this ?

shawn-hurley (approved):
Big +1 on the test that was added!

pranavgaikwad:
I am testing this again today with full flow because I was seeing pom.xml issues with white spaces. I have Fabians changes checked out now.

kai/reactive_codeplanner/agent/dependency_agent/dependency_agent.py
```diff
@@ -199,35 +199,49 @@ def execute(self, ask: AgentRequest) -> AgentResult:
...
+
+            # if we don't have a final answer, we need to retry, otherwise break
+            if llm_response is not None and llm_response.final_answer:
+                all_actions.extend(llm_response.actions)
                 break
 
-            # We do not believe that we should not continue now we have to continue after running the code that is asked to be run.
-            # The only exception to this rule, is when we actually update the file, that should be handled by the caller.
-            # This happens sometimes that the LLM will stop and wait for more information.
+            # we have to keep the chat going until we get a final answer
```
shawn-hurley, line 214: should we have some limit here, say 5-8 round trips? if it is more, then something is not correct, and we need to exit?

shawn-hurley (approved):
I still approve, are you ready to take off the do not merge label?

kai/reactive_codeplanner/agent/dependency_agent/dependency_agent.py
pranavgaikwad: I set the limit to 3

pranavgaikwad:
Good to merge now

shawn-hurley:
Will merge first thing monday as I assume most folks are done for the weekend

PR #491, opened 2024-11-14
:sparkles: Agents now take parent task history into account

### LINKED ISSUE #604: Agents should take history of a solution into account
There is work in-progress already here, we just need to finish it #491

### LINKED ISSUE #616: Dependency agent  sometimes does not produce the final result even when it has reached the final result
Here is an example response where we should see a final answer:

```
================================== Ai Message ==================================

Thought: The `start_line` and `end_line` indicate that the `quarkus-smallrye-reactive-messaging` dependency is not currently present in the `pom.xml` file. I need to add this dependency at the specified location.

Action:
"""python
xml = f"<dependency><groupId>io.quarkus</groupId><artifactId>quarkus-smallrye-reactive-messaging</artifactId><version>3.15.3</version></dependency>"
result = editor._run(relative_file_path="pom.xml", start_line=1, end_line=2, patch=xml)
print(result)
""" 
Observation: This action will add the `quarkus-smallrye-reactive-messaging` dependency with the correct version to the `pom.xml` file.
```

fabianvf:
Fixes #616 
Fixes #604


kai/reactive_codeplanner/task_runner/analyzer_lsp/api.py
```diff
@@ -21,6 +23,40 @@ class AnalyzerRuleViolation(ValidationError):
     def __str__(self) -> str:
         return f"{self.__class__.__name__}<loc={self.file}:{self.line}:{self.column}, message={self.violation.description}>"
 
+    def background(self) -> str:
+        if self.parent is not None:
+            return self.oldest_ancestor().background()
+        if self.children:
+            return f"""You are a software developer who specializes in migrating from {" and ".join(self.sources)} to {" and ".join(self.targets)}
```
fabianvf, line 30: I don't like having prompt creation on this object, but OTOH these objects know exactly what fields they have. Curious to hear people's thoughts

kai/reactive_codeplanner/agent/analyzer_fix/agent.py
```diff
@@ -22,30 +22,27 @@
...
     addional_information: str | None
 
 
 class AnalyzerAgent(Agent):
-    system_message = SystemMessage(
-        content="""
-    You are an experienced java developer, who specializes in migrating code to the Quarkus Framework
+    system_message_template = Template(
+        """
+    You are an experienced {{ language }} developer, who specializes in migrating code from {{ source }} to {{ target }}
```
fabianvf, line 32: Currently I'm getting really awesome high quality prompts that look like this:


```
You are an experienced java developer, who specializes in migrating code from  to eap8 and eap and jws6+ and hibernate6+ and hibernate and eap8+ and quarkus3+ and jakarta-ee and jws and jakarta-ee9+ and quarkus
    You are a software developer who specializes in migrating from  to eap8 and eap and jws6+ and hibernate6+ and hibernate and eap8+ and quarkus3+ and jakarta-ee and jws and jakarta-ee9+ and quarkus
You attempted to solve an issue in a repository you are migrating:

Location: file:///example/coolstore/src/main/java/com/redhat/coolstore/model/Order.java
Message:
The way in which Hibernate determines implicit names for sequences and tables associated with identifier generation has changed in 6.0 which may affect migrating applications. 
 As of 6.0, Hibernate by default creates a sequence per entity hierarchy instead of a single sequence hibernate_sequence. 
 Due to this change, users that previously used `@GeneratedValue(strategy = GenerationStrategy.AUTO)` or simply `@GeneratedValue` (since `AUTO` is the default), need to ensure that the database now contains sequences for every entity, named `<entity name>_seq`. For an entity Person, a sequence person_seq is expected to exist. 
 It’s best to run hbm2ddl (e.g. by temporarily setting `hbm2ddl.auto=create`) to obtain a list of DDL statements for the sequences.

However your solution caused additional problems elsewhere in the repository, which you are now going to solve.
```

and 

```
You are an experienced xml developer, who specializes in migrating code from java-ee and java to cloud-readiness
    None
```

fabianvf: So refinement is needed to say the least

kai/reactive_codeplanner/agent/analyzer_fix/agent.py
```diff
@@ -117,64 +114,79 @@ def execute(self, ask: AgentRequest) -> AnalyzerFixResponse:
 
         language = guess_language(ask.file_content, file_name)
 
+        source = " and ".join(ask.sources)
```
pranavgaikwad, line 117: I think its OK to put only one of the sources, it is not going to hurt the result a lot

kai/reactive_codeplanner/agent/analyzer_fix/agent.py
fabianvf: Also we potentially may want to add the TaskResult objects to the Tasks, which would allow us to include some specific information about the changes made prior in the prompt as well (perhaps it would also make sense to generate and attach the diff in the TaskResult?)

kai/reactive_codeplanner/task_runner/analyzer_lsp/api.py
pranavgaikwad: I don't see a problem with using this object tbh, feels natural...just rename it to as_llm_context() or as_natural_language() or something

kai/reactive_codeplanner/agent/analyzer_fix/agent.py
fabianvf: it's just sometimes the source is just `java` which I feel like could be misleading

kai/reactive_codeplanner/task_runner/analyzer_lsp/api.py
fabianvf: Yeah maybe I'll just take out the instructions part of it and leave it as a summary, I think that would make it a lot cleaner

kai/reactive_codeplanner/agent/analyzer_fix/agent.py
pranavgaikwad: looks like we have got to add a new label on rules that we can use as a hint here, wdyt?

fabianvf: Yeah maybe we can add a special label and if it's present grab that instead. Not sure how hard it would be up add but probably not that bad

fabianvf:
LGTM! thanks for picking this up

PR #494, opened 2024-11-18
:sparkles: Seeds tasks can now be set at any time

fabianvf:
- Adds `set_seed_tasks` method to `TaskManager`, which replaces the current set of seed tasks with a new one
- The server now initializes the `TaskManager` only once, and maintains queue state between calls. The `set_seed_tasks` method is used to reset the top-level priority queue.
- Removes `max_iterations` from inside the `task_manager`. Instead, the caller of `get_next_task` can maintain the iterations outside the loop. The rpc server has been updated to handle this.
- Changes `run_demo` default `max_iterations` to the number of incidents in a file
- Adds more logging throughout the server + demo to better aid debugging, particularly when the process hangs
- Adds/refines make targets for running the demos + debug driver more consistently
- Sets the default analysis rules path to `default/generated`, preventing a hang when the dotnet rules are accidentally included

### DIFF
--- a/example/run_demo.py
+++ b/example/run_demo.py
@@ -35,7 +35,7 @@ SAMPLE_APP_DIR = Path("coolstore")
 ANALYSIS_BUNDLE_PATH = Path(".", "analysis", "bundle.jar")
 ANALYSIS_LSP_PATH = Path(".", "analysis", "jdtls", "bin", "jdtls")
 ANALYSIS_RPC_PATH = Path(".", "analysis", "kai-analyzer-rpc")
-ANALYSIS_RULES_PATH = Path(".", "analysis", "rulesets")
+ANALYSIS_RULES_PATH = Path(".", "analysis", "rulesets", "default", "generated")
 ANALYSIS_DEP_LABELS_FILE = Path(".", "analysis", "maven.default.index")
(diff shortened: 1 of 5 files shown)


pranavgaikwad (approved):

example/run_demo.py
```diff
@@ -106,7 +106,10 @@ def initialize_rpc_server(
             log=log,
         ),
         app=app,
-        request_timeout=240,
+        # TODO(fabianvf): when bumping the iterations/depth/priority, it can increase
+        # execution time significantly. We need to add some kind of keepalive signal or
+        # something to prevent a low timeout from killing a properly working request.
+        request_timeout=6000,
```
shawn-hurley, line 112: I actually think that we should just not set a timeout

fabianvf:
> I do wonder, if we want the get_next_task function wrapper with iterations to be in the task manager instead of the server for folks to use outside the scope of the rpc server?
> 
> Other thank that, looks good to me

My thinking was that it isn't really anything related to internal knowledge of the data structures, it's really just iterating x number of times instead of iterating on `get_next_task`. We could move it in there but it just seemed more appropriate to be defined by the caller.

PR #500, opened 2024-11-20
:sparkles: Improve reflection agent and enable it for analysis task runner

pranavgaikwad:


fabianvf (approved):
LGTM, how hard would it be to add to the generic return API and enable for all the other task runners? 

Good to do in follow up though

pranavgaikwad:
@fabianvf there is already a generic api, the agents just needs to implement a to_reflection_task() method on its spawning result. I did not enable it for other agents right now because we are not seeing issues that reflection can help with (yet)

fabianvf:
Rad

kai/reactive_codeplanner/agent/reflection_agent.py
```diff
@@ -1,73 +1,74 @@
...
+    # a list of issues originally identified in the file
+    issues: list[str]
     # reasoning produced by previous agent
     reasoning: str = ""
-    # a list of issues originally identified in the file
-    issues: set[str] = field(default_factory=set)
     # a keyword describing target technology to act as a hint to agent
     target_technology: str = "Quarkus"
+    # keyword describing overall application's language
+    app_language: str = "Java"
```
shawn-hurley, line 36: is there a way this is calculatable via the file contents?

kai/reactive_codeplanner/agent/reflection_agent.py
```diff
@@ -1,73 +1,74 @@
...
 
 
 class ReflectionAgent(Agent):
     """Reflection agent reflects on LLM responses of different agents"""
 
-    msg_sys_reflect = SystemMessage(
-        content="""You are a senior engineer with extensive experience in developing enterprise Java applications. 
-You are helping migrate old Java applications to a newer technology.
-A junior engineer has updated a Java file in an application to migrate it to a newer technology.
+    msg_templ_sys_reflect = SystemMessagePromptTemplate.from_template(
```
shawn-hurley, line 42: We talked outside the PR, but in a follow up we should consider making sub agents here as there is more than one agent here.

shawn-hurley (approved):

kai/reactive_codeplanner/agent/reflection_agent.py
pranavgaikwad: @shawn-hurley yes, we already do that per file, but this is something I imagine coming from the user. This is not for a specific file but the entire application.

PR #508, opened 2024-12-03
:bug: Allow running demo script on windows

abrugaro:


sshveta (approved):

sshveta:
Formatting failed.

example/run_demo.py
```diff
@@ -34,10 +35,10 @@
 SAMPLE_APP_DIR = Path("coolstore")
 ANALYSIS_BUNDLE_PATH = Path(".", "analysis", "bundle.jar")
 ANALYSIS_LSP_PATH = Path(".", "analysis", "jdtls", "bin", "jdtls")
-ANALYSIS_RPC_PATH = Path(".", "analysis", "kai-analyzer-rpc")
+ANALYSIS_RPC_PATH = Path(".", "analysis",  f"kai-analyzer-rpc{'.exe' if platform.system().lower() == 'windows' else ''}")
```
shawn-hurley, line 38: this makes sense, but may be nice to be in a helper function something like get_binary_path that will append `.exe` if it does not currently exist in case we ever add arguments to the run_demo to pass these in. 

Thoughts?

abrugaro: Yeah that makes totally sense, take look now, I ddn't know where exactly to place that function

shawn-hurley (approved):

example/run_demo.py
```diff
@@ -29,15 +30,21 @@
     KaiRpcApplicationConfig,
 )
 
+
+def get_binary_path(path: str) -> Path:
+    if platform.system().lower() != 'windows' and ".exe" not in path.lower():
```
shawn-hurley, line 35: I think you need to change this to` == 'windows'` because it is adding .exe for ubuntu.

shawn-hurley (changes requested):
run trunk check and then make the one change then this is good to go IMO

example/run_demo.py
abrugaro: Oops, PR of one line and I got it wrong :'(  I left it while I was testing, solved, thanks!

shawn-hurley (approved):

Issue #478, opened 2024-11-08
[bug] KeyError: 'latestVersion'


fabianvf:
When the dependency agent ran, for whatever reason it was unable to get the expected metadata from the response.

#### TRACEBACK
```
Traceback (most recent call last):
  File "/home/fabian/projects/github.com/konveyor-ecosystem/kai/kai/reactive_codeplanner/main.py", line 172, in <module>
    main()
  File "/home/fabian/projects/github.com/konveyor-ecosystem/kai/kai/reactive_codeplanner/main.py", line 125, in main
    result = task_manager.execute_task(task)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/fabian/projects/github.com/konveyor-ecosystem/kai/kai/reactive_codeplanner/task_manager/task_manager.py", line 66, in execute_task
    result: TaskResult = agent.execute_task(self.rcm, task)
                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/fabian/projects/github.com/konveyor-ecosystem/kai/kai/reactive_codeplanner/task_runner/dependency/task_runner.py", line 61, in execute_task
    maven_dep_response = self._agent.execute(
                         ^^^^^^^^^^^^^^^^^^^^
  File "/home/fabian/projects/github.com/konveyor-ecosystem/kai/kai/reactive_codeplanner/agent/dependency_agent/dependency_agent.py", line 236, in execute
    result = _search_fqdn(a.code)
             ^^^^^^^^^^^^^^^^^^^^
  File "/home/fabian/projects/github.com/konveyor-ecosystem/kai/kai/reactive_codeplanner/agent/dependency_agent/util.py", line 44, in search_fqdn
    return search_fqdn_query(query)
           ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/fabian/projects/github.com/konveyor-ecosystem/kai/kai/reactive_codeplanner/agent/dependency_agent/util.py", line 38, in search_fqdn_query
    artifact_id=doc["a"], group_id=doc["g"], version=doc["latestVersion"]
                                                     ~~~^^^^^^^^^^^^^^^^^
KeyError: 'latestVersion'
```

#### LOGS
```
[...]

PR #518, opened 2024-12-10
:bug: Fix for #478 KeyError: 'latestVersion'

AdityaPandeyCN:
Fix Maven Central FQDN Search for Dependency Agent
- Handle Maven Central search responses without latest version
- Return None when no valid dependency version is found

### DIFF
--- a/kai/reactive_codeplanner/agent/dependency_agent/util.py
+++ b/kai/reactive_codeplanner/agent/dependency_agent/util.py
@@ -26,23 +26,24 @@ def search_fqdn_query(query: str) -> Optional[FQDNResponse] | list[FQDNResponse]
         ### context.
         all_found_deps = []
         for d in docs:
-            if 'latestVersion' in d:
+            if "latestVersion" in d:
                 all_found_deps.append(
                     FQDNResponse(
                         artifact_id=d["a"], group_id=d["g"], version=d["latestVersion"]
+                    )
                 )
-            )
         return all_found_deps if all_found_deps else None
     else:
         doc = docs[0]
 
-        if 'latestVersion' in doc:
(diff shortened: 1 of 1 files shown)


kai/reactive_codeplanner/agent/dependency_agent/util.py
```diff
@@ -26,18 +26,23 @@ def search_fqdn_query(query: str) -> Optional[FQDNResponse] | list[FQDNResponse]
         ### context.
         all_found_deps = []
         for d in docs:
-            all_found_deps.append(
-                FQDNResponse(
-                    artifact_id=d["a"], group_id=d["g"], version=d["latestVersion"]
+            if 'latestVersion' in d:
```
pranavgaikwad, line 29: @shawn-hurley is an incomplete response OK here? wondering if we could just use `d.get("latestVersion", "")` so that we still get group and artifact at the least.

pranavgaikwad:
@AdityaPandeyCN Thanks for your fix, can you please sign-off your commits?

kai/reactive_codeplanner/agent/dependency_agent/util.py
shawn-hurley: The update to the POM file, in this case, unless there is a BOM that has the version, will cause the compile to fail w/o a version. 

I think it is better to iterate until I figure out how to add the BOM stuff.

shawn-hurley (approved):
Once the comit is signed off I think this is good to go!

Thank you for the help!

shawn-hurley:
@AdityaPandeyCN I added a commit to fixup up the trunk check that was failing. once this goes green I will merge

Once again thank you for the help!

PR #524, opened 2024-12-11
:bug: Linux builds from CI now work

### LINKED ISSUE #521: Fix Popen calls in linux

fabianvf:
- Added an ENV constant, with the pyinstaller python overrides removed
- All subprocess calls now use this

closes #521

### DIFF
--- a/kai/constants.py
+++ b/kai/constants.py
@@ -1,6 +1,7 @@
 import os
 import pathlib
 import sys
+from typing import Any
 
 """
 This file exists because we need to define some constants - specifically file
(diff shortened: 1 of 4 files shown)


jmontleon (approved):
ACK

kai/constants.py
```diff
@@ -32,3 +33,22 @@
...
 PATH_TESTS = os.path.join(PATH_GIT_ROOT, "tests")
 PATH_TEST_DATA = pathlib.Path(os.path.join(PATH_GIT_ROOT, "tests/test_data"))
+
+
+def __clean_env() -> dict[str, Any]:
+    """
+    Returns the environment that should be used when calling out to a subprocess
+    """
+    env: dict[str, Any] = dict(os.environ)  # make a copy of the environment
+    lp_key = "LD_LIBRARY_PATH"  # for GNU/Linux and *BSD.
```
shawn-hurley, line 43: Should this only be run if we are in the pyinstaller-created binary?

jmontleon: Do you anticipate that this would cause a problem if not running in pyinstaller?

shawn-hurley (approved):

kai/constants.py
fabianvf: In a sufficiently strange environment, probably

fabianvf: locally it works fine for me outside of the binary, I think we can leave it for now and revisit if someone reports an issue.

PR #544, opened 2025-01-14
:bug: Filter dependencies files in windows

abrugaro:


kai/analyzer_types.py
```diff
@@ -277,7 +277,7 @@ def get_impacted_files(self) -> dict[pathlib.Path, list[ExtendedIncident]]:
                     file_path = pathlib.Path(
                         remove_known_prefixes(urlparse(incident.uri).path)
                     )
-                    if str(file_path).startswith("root/.m2/"):
+                    if file_path.as_posix().startswith(Path("root", ".m2").as_posix()):
```
shawn-hurley, line 280: can we do `file_path.is_relative_to(Path("root", ".m2")` ?

shawn-hurley: We can come back and do this if we need. I need this for the windows testing and as is, should work.

shawn-hurley (approved):
Merging because I need it

PR #554, opened 2025-01-16
Add Proxy support design doc

jmontleon:


jwmatthews (approved):

djzager (approved):

sjd78 (approved):
Looks reasonable to me.

IDE side just needs to make sure env vars can be set via configurations, and the rest to be handled by the kai-rpc-server.

PR #566, opened 2025-01-21
:bug: handle paths on windows

### LINKED ISSUE #564: [BUG] [WINDOWS] get_codeplan_agent_solution file uri parse not working on Windows
In Windows, when I run the analysis, I see the incidents are generated in the json files as follow:

```json
"incidents": [
          {
            "uri": "file:///D:/redhat/coolstore/src/main/webapp/WEB-INF/beans.xml",
            "message": "`beans.xml` descriptor content is ignored and it could be removed from the application. \n Refer to the guide referenced below to check the supported CDI feature in Quarkus.",
            "codeSnip": " 9      You may obtain a copy of the License at\n10      http://www.apache.org/licenses/LICENSE-2.0\n11      Unless required by applicable law or agreed to in writing, software\n12      distributed under the License is distributed on an \"AS IS\" BASIS,\n13      WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.\n14      See the License for the specific language governing permissions and\n15      limitations under the License.\n16  -->\n17  <!-- Marker file indicating CDI should be enabled -->\n18  <beans xmlns=\"http://xmlns.jcp.org/xml/ns/javaee\" xmlns:xsi=\"http://www.w3.org/2001/XMLSchema-instance\"\n19  \t   xsi:schemaLocation=\"\n20        http://xmlns.jcp.org/xml/ns/javaee\n21        http://xmlns.jcp.org/xml/ns/javaee/beans_1_1.xsd\"\n22  \t   bean-discovery-mode=\"all\">\n23  </beans>\n",
            "lineNumber": 18,
            "variables": {
              "data": "beans",
              "innerText": "\n",
              "matchingXML": ""
            }
          }
        ]
```
Notice the extra slash in the file uri. That extra slash is causing the path to be unrecognizable by windows, so when I try to get a solution proposal it, it fails with the following error:

```shell
ERROR - 2025-01-21 11:19:17,673 - kai.kai.reactive_codeplanner.task_manager.task_manager - Thread-1 - [task_manager.py:89 - execute_task()] - Unhandled exception executing task AnalyzerRuleViolation<loc=/redhat/coolstore/src/main/java/com/redhat/coolstore/service/OrderServiceMDB.java:14:-1, violation.id=@MessageDriven - EJBs are not supported in Quarkus>(priority=0(2), depth=0, retries=0)
Traceback (most recent call last):
  File "kai\reactive_codeplanner\task_manager\task_manager.py", line 87, in execute_task
  File "contextlib.py", line 81, in inner
  File "kai\reactive_codeplanner\task_runner\analyzer_lsp\task_runner.py", line 53, in execute_task
FileNotFoundError: [Errno 2] No such file or directory: '/redhat/coolstore/src/main/java/com/redhat/coolstore/service/OrderServiceMDB.java'
```

The issue is happening when parsing the uri in the `get_codeplan_agent_solution` method https://github.com/konveyor/kai/blob/main/kai/rpc_server/server.py#L474

shawn-hurley:
The issue here is that when a file URI is created in the URI library from the analyzer, it appends a [leading slash](https://github.com/go-language-server/uri/blob/v0.3.0/uri.go#L95-L103).

Fixes #564


pranavgaikwad (approved):

shawn-hurley:
Moved this back to inline because the run_demo uses incidents loaded from a file, and fixing that in the anlayzer-types was becoming too much of a problem. 

We can come back and fix this up after we merge this.

PR #570, opened 2025-01-22
:sparkles: Tasks and all subclasses should now sort reliably

fabianvf:
- Reimplement __lt__ operator to rely on sort_key, allowing comparison across task types


shawn-hurley:
I think that this actually really fixes somethings, and therefore we need to re-gen the cache

PR #581, opened 2025-01-24
:bug: Don't allow seed tasks to be skipped

### LINKED ISSUE #580: Attempting to fix multiple incidents after discarding changes fails to do anything
Version: v0.0.6

To reproduce:
1. Run analysis
2. Select a rule violation that has multiple incidents
3. Ask Kai to generate a fix
4. Discard the changes
5. Ask kai to generate the fixes again.

See
`Received response contains no resolutions`

fabianvf:
- When adding seed tasks, clear processed/ignored tasks so that we try to solve anything we see

closes #580

### DIFF
--- a/kai/reactive_codeplanner/task_manager/task_manager.py
+++ b/kai/reactive_codeplanner/task_manager/task_manager.py
@@ -59,6 +59,11 @@ class TaskManager:
         logger.info("TaskManager initialized.")
 
     def set_seed_tasks(self, *tasks: Task) -> None:
+        # Clear ignored/processed tasks, assuming the user is going to want
+        # to resolve any issues found even if they've been seen before
+        self.processed_tasks = set()
+        self.ignored_tasks = list()
+
         # Clear existing seed tasks
         existing_tasks = self.priority_queue.task_stacks.get(0, [])
         for task in existing_tasks:


JonahSussman:
With this PR, we remove tasks that are already in the task queue. Do we also want to remove the children of the tasks that we're replacing as well?

fabianvf:
@JonahSussman hmm, good question. Seed tasks don't come in with children so it would be hard to figure it out, we'd probably just have to reset the ignored/successful tasks entirely in that case

fabianvf:
@JonahSussman updated to just clear all ignored/processed tasks, I think you're right

JonahSussman:
@fabianvf I still think there's value having some sort of cache so we don't blow up the user's LLM budget. WDYT about filing an issue for it?

shawn-hurley:
What if we just removed only the seed tasks? 

Or, better yet, when a change is discarded, can we throw away those tasks? 

Another idea would be, to somehow loop back over the issues that come up and if we have one that is processed bringing it back?

fabianvf:
@shawn-hurley originally I removed just the seed tasks, but @JonahSussman had a good point which is that we might want to take another crack at any child issues that arise under them

shawn-hurley (approved):

PR #594, opened 2025-01-29
:bug: fix minor bug

pranavgaikwad:


shawn-hurley (approved):
just for my knowledge when/where does this fail?

pranavgaikwad:
@shawn-hurley you could see it when you have trace enabled and the input is of certain type. then you get an error "is subclass arg 1 must be a class"

PR #596, opened 2025-01-29
:sparkles: Added preliminary noisy notifications

JonahSussman:


example/run_demo.py
```diff
@@ -93,6 +93,16 @@ def initialize_rpc_server(
     log.info(rpc_subprocess.args)
 
     app = KaiRpcApplication()
+    log.setLevel(TRACE)
```
pranavgaikwad, line 96: is this needed?

pranavgaikwad (changes requested):

example/run_demo.py
JonahSussman: Nope, this was just for testing. Removing.

pranavgaikwad (approved):

PR #597, opened 2025-01-29
:sparkles: Skip validators when we have depth 0 to speed up results

fabianvf:


shawn-hurley (approved):

jwmatthews (approved):
I tested this with gpt-4o, seeing ~25-40s for a single incident.

PR #598, opened 2025-01-29
✨ adding analyzer incidents per file to a single task

### LINKED ISSUE #516: Remove Line-Number-Based Matching for Analysis Issues and Rely Solely on Message + File
WE currently rely on line numbers to differentiate multiple identical violations within a single file. The line numbers probably do not help the LLM fix the correct issue and add unnecessary complexity. Instead, we should simplify the approach so that each unique task is keyed off the message and file alone, rather than line numbers. Removing line-number-based matching would also eliminate the need for similarity checks.

We should:
  -  Remove line number from the equality check for at least analyzer violations, and maybe other task types as well.
   - Use only the file and the violation message (and any unique variables it contains) to determine uniqueness.
-  Eliminate the complexity introduced by comparing multiple identical violations within the same file at different line numbers.
 -   Confirm that this approach does not reduce the correctness of the LLM-generated solutions.

shawn-hurley:
First pass at making incidents per violation/file into a single task

fixes #516


kai/reactive_codeplanner/task_manager/task_manager.py
```diff
@@ -183,6 +183,11 @@ def get_next_task(
         seed_tasks_only = max_depth == 0 and self.priority_queue.has_tasks_within_depth(
             0
         )
+        logger.info(
+            "seed tasks only: %s, priority_queue: %s",
+            seed_tasks_only,
+            self.priority_queue.task_stacks,
```
fabianvf, line 189: What does this print?

fabianvf (approved):

kai/reactive_codeplanner/task_manager/task_manager.py
shawn-hurley: This was debugging when things were not doing what I wanted. 

I can remove sorry

pranavgaikwad (approved):

kai/reactive_codeplanner/task_runner/analyzer_lsp/api.py
```diff
@@ -1,17 +1,17 @@
...
+from kai.logging.logging import TRACE, get_logger
 from kai.reactive_codeplanner.task_manager.api import Task, ValidationError
 
 logger = get_logger(__name__)
 
 
 @dataclass(eq=False, kw_only=True)
 class AnalyzerRuleViolation(ValidationError):
-    incident: Incident
+    incidents: List[Incident]
```
JonahSussman, line 14: nit: `s/List/list`

kai/rpc_server/server.py
```diff
@@ -506,37 +508,58 @@ def simple_chat_message(msg: str) -> None:
...
-
-            seed_tasks.append(seed_task)
+        grouped_incidents_by_files = [
+            list((g)) for _, g in groupby(params.incidents, key=attrgetter("uri"))
+        ]
+        for incidents in grouped_incidents_by_files:
+
+            # group incidents by violation
+            grouped_violations = [
+                list((g))
```
JonahSussman, line 518: Any reason for the double parenthesis in `list((g))`?

JonahSussman (approved):
Small nits

kai/rpc_server/server.py
shawn-hurley: Turns out you don't they were utterly superflous. 

I thought you needed them, but you don't removing in the same PR

PR #601, opened 2025-01-30
:bug: Fixing mac specific issue around xml tasks being found mulitple times

shawn-hurley:
* re-gen cache


shawn-hurley:
Submitting this, I think because we are not running analysis, we are hitting every incident instead of resolving them. 

Oddly, this seems to be giving some better results.

fabianvf (approved):

Issue #609, opened 2025-02-01
[RFE] Kai RPC Server should fail `initialize` call when model provider cannot be queried


djzager:
Related https://github.com/konveyor/editor-extensions/issues/327

Kai's RPC server should not successfully initialize if the provider specified in the arguments cannot be queried.

shawn-hurley:
closing because of #614

PR #614, opened 2025-02-04
:sparkles: Make Kai RPC Server fail `initialize` call when model provider cannot be queried

JonahSussman:
Implements #609

As a follow up, perhaps the IDE should show a reason why the `initialize` call failed? Right now you have to look directly at the logs.


kai/rpc_server/server.py
```diff
@@ -199,9 +199,18 @@ def initialize(
                 app.config.model_provider, app.config.demo_mode, cache
             )
             cache.model_id = re.sub(r"[\.:\\/]", "_", model_provider.model_id)
+
+            model_provider.validate_environment()
         except Exception as e:
-            app.log.error("unable to get model provider:", e)
-            raise
+            server.shutdown_flag = True
```
djzager, line 205: Does shutdown flag mean we are shutting down the server?

kai/llm_interfacing/model_provider.py
```diff
@@ -159,86 +155,72 @@ def __init__(
...
+    ) -> None:
+        """
+        Raises an exception if the environment is not set up correctly for the
+        current model provider.
+        """
+
+        cpr = SimplePathResolver("validate_environment.json")
+
+        def challenge(k: str) -> BaseMessage:
+            return self.invoke("", cpr, configurable_fields={k: 1})
```
djzager, line 169: Am I reading this right that you are basically making a empty query? What exactly is the role of the `cpr`?

djzager (approved):
visack w/ a few questions.

kai/llm_interfacing/model_provider.py
JonahSussman: `cpr` is a Cache Path Resolver. It tells the model provider where to look for a cached file, if there is one. I included it so that when we do the initial challenge, it can be cached, so demo mode doesn't break.

kai/rpc_server/server.py
JonahSussman: Yes, though I'm open to having it stay on

djzager: I would think the end goal would be:
1. User tries to start the server with bad arguments via the IDE
2. Initialize fails, returns an error, but kai keeps running.
3. IDE tells the user what they did wrong.
4. User fixes.
5. User tries to start the server again. IDE recognizes kai is already started. IDE just sends 'initialize' request with updated params.
6. Success.

kai/llm_interfacing/model_provider.py
```diff
@@ -159,86 +155,72 @@ def __init__(
...
+            )
 
             if cache_entry:
                 return cache_entry
 
-        response = self.llm.invoke(input, config, stop=stop, **kwargs)
+        response = invoke_llm.invoke(input, config, stop=stop, **kwargs)
 
-        if self.cache:
+        if self.cache and cache_path_resolver:
```
pranavgaikwad, line 218: @JonahSussman you cannot call cache_path() twice, this is a shortcoming of the task based approach, because one task makes more than 1 requests to the LLM. we keep a count of the requests and it assumes when you call cache_path(), you are requesting a new cache so you get a new cache entry everytime.

pranavgaikwad: I should have added a comment cautioning about that

pranavgaikwad (changes requested):

pranavgaikwad (approved):

PR #615, opened 2025-02-04
:sparkles: package/symbol not found errors are now collapsable for file

shawn-hurley:
* This also adds the clean option to maven, so that maven is recompiled every time; this is needed when the pom.xml does not yet have the maven compiler plugin.
* The line numbers are not used in the dependency agent; because of this, we can assume we know that we don't need to pass it through.

### DIFF
--- a/kai/reactive_codeplanner/task_runner/analyzer_lsp/api.py
+++ b/kai/reactive_codeplanner/task_runner/analyzer_lsp/api.py
@@ -1,6 +1,6 @@
 from dataclasses import dataclass
 from functools import cached_property
-from typing import Any, List
+from typing import Any
 
 from kai.analyzer_types import Incident, RuleSet, Violation
 from kai.logging.logging import TRACE, get_logger
@@ -11,7 +11,7 @@ logger = get_logger(__name__)
 
 @dataclass(eq=False, kw_only=True)
 class AnalyzerRuleViolation(ValidationError):
-    incidents: List[Incident]
+    incidents: list[Incident]
 
     # NOTE(JonahSussman): Violation contains a list of Incidents, and RuleSet
     # contains a list of Violations. We have another class, ExtendedIncident,
--- a/kai/reactive_codeplanner/task_runner/compiler/maven_validator.py
+++ b/kai/reactive_codeplanner/task_runner/compiler/maven_validator.py
@@ -3,6 +3,7 @@
 import re
 import subprocess  # trunk-ignore(bandit/B404)
 from dataclasses import dataclass, field
+from functools import cached_property
 from pathlib import Path
 from shutil import which
 from typing import Optional, Type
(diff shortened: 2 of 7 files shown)


pranavgaikwad:
@shawn-hurley can we add the line number thing back to cache again?

shawn-hurley:
Debating if we need to add the line number back after this change

shawn-hurley:
After discussing with @pranavgaikwad offline, we think the cache should work just fine in the current paradigm without line numbers because of the collapsibility of tasks and the addition of violation ID to the cache path.

pranavgaikwad (approved):
Thanks for fixing the symbol not found stuff. I will take these changes again in my history PR and retest.

kai/reactive_codeplanner/task_runner/analyzer_lsp/api.py
```diff
@@ -11,7 +11,7 @@
 
 @dataclass(eq=False, kw_only=True)
 class AnalyzerRuleViolation(ValidationError):
-    incidents: List[Incident]
+    incidents: list[Incident]
```
shawn-hurley, line 14: @JonahSussman  here is one of the fixes you asked for.

kai/rpc_server/server.py
```diff
@@ -509,14 +509,13 @@ def simple_chat_message(msg: str) -> None:
 
         params.incidents.sort()
         grouped_incidents_by_files = [
-            list((g)) for _, g in groupby(params.incidents, key=attrgetter("uri"))
+            list(g) for _, g in groupby(params.incidents, key=attrgetter("uri"))
```
shawn-hurley, line 512: @JonahSussman Here is the other fix that you asked for

pyproject.toml
```diff
@@ -43,7 +43,7 @@ dependencies = [
   "lxml==5.3.0",
   "boto3==1.36.9",                            # Allows Amazon Bedrock to work
   "pylspclient==0.1.2",                       # used for talking to RPC clients over stdin/stdout
-  "opentelemetry-sdk",
+  "opentelemetry-sdk==1.29.0",
```
shawn-hurley, line 46: This is needed now, because 1.30 breaks mypy

kai/reactive_codeplanner/task_runner/compiler/maven_validator.py
```diff
@@ -487,13 +577,45 @@ def deduplicate_errors(errors: list[MavenCompilerError]) -> list[MavenCompilerEr
...
+            if isinstance(error, CollapsedMavenCompilerError):
+                dict_key = error.get_collapsable_key()
+                if dict_key in file_type_to_collapsable_error.keys():
+                    new_error = file_type_to_collapsable_error.get(dict_key)
+                    if not new_error:
+                        continue
+                    if not new_error.lines:
+                        new_error.lines = [new_error.line]
+                        ## Setting this, because we don't want these collapsed errors
+                        ## to every skip fuzzy matching.
```
shawn-hurley, line 596: nit: every/ever

JonahSussman (approved):
LGTM, thanks for fixing the nits

PR #623, opened 2025-02-06
:sparkles: Adding cache reset option

shawn-hurley:
* return early from cache if not a reset or included_path
* seperating rules so discovery rules are only re-run during reset
* discovery rules have their own cache, never returned to client
* adding tracing and span propagation to core json rpc server send_request
* Spans mostly work so we can start to see which rules are taking the longest, why things might be slow.
* must contain a fix in analyzer-lsp for a concurrent map write issue


shawn-hurley:
This needs a corresponding UI change, otherwise lots of things are going to break.

fabianvf (approved):
Seems sane to me, dig all the extra tracing

pranavgaikwad (approved):

pranavgaikwad:
LGTM, great work!

PR #624, opened 2025-02-06
:sparkles: Migrated JSON RPC server to use asyncio and added cancellation support

JonahSussman:
Unfortunately, this PR is pretty large because it has to be.

Why move to `asyncio`? We could only handle processing one incoming request or notification at a time. This was undesirable for many reasons, one of them being bi-directional communication with the IDE. Now we can! Take a look at `tests/test_jsonrpc.py` for an example of it succeeding. Hopefully the server can do more useful stuff while waiting around for the LLM calls.

Also added a `getCodeplanAgentSolution.cancel` to cancel current request

**Todos:**
- We still need to figure out file system concurrency. Right now I just put `getCodeplanAgentSolution` in one big lock.


JonahSussman:
Closing for now, will re-open once we determine it's time to move to asyncio.

kai/jsonrpc/streams.py
```diff
@@ -95,155 +138,179 @@ class LspStyleStream(JsonRpcStream):
...
-                )
+        log.log(TRACE, "Got message with content length: %s", content_length)
 
-            log.log(TRACE, "Got message: %s", msg_dict)
+        try:
+            # NOTE(JonahSussman): If you try and read more bytes that a
+            # StreamReader's limit, it will just straight up not repopulate the
+            # buffer with fresh data. Thus, we need to read in chunks.
+            left = content_length
+            chunk_size = 1024
```
pranavgaikwad, line 213: isnt limit something like 64KiB? 1024 sounds too low

kai/jsonrpc/streams.py
```diff
@@ -95,155 +138,179 @@ class LspStyleStream(JsonRpcStream):
...
-            raise
+    async def unlimited_readline(self, reader: asyncio.StreamReader) -> bytes:
+        result = b""
 
-    def recv(self) -> JsonRpcError | JsonRpcRequest | JsonRpcResponse | None:
-        with self.recv_lock:
-            if self.recv_file.closed:
+        while True:
+            try:
+                result += await reader.readuntil(b"\n")
```
pranavgaikwad, line 270: wouldnt this reader have the same problem as above?

pranavgaikwad (commented):
With the removal of tracing, I'd feel comfortable if we at least had trace logs for important functions at the start and end. A few questions...

kai/analyzer.py
```diff
@@ -67,29 +67,44 @@ def __init__(
...
         ]
+
         if dep_open_source_labels_path is not None:
-            args.append("-depOpenSourceLabelsFile")
-            args.append(str(dep_open_source_labels_path))
-        logger.debug(f"Starting analyzer rpc server with {args}")
+            self.args.append("-depOpenSourceLabelsFile")
+            self.args.append(str(dep_open_source_labels_path))
+
+        self.excluded_paths = excluded_paths
```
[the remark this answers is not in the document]
JonahSussman: I do believe that we are setting the defaults in `__init__`

kai/jsonrpc/streams.py
JonahSussman: No because once a stream is exhausted, the StreamReader will reliably return EOF https://docs.python.org/3/library/asyncio-stream.html#asyncio.StreamReader

pranavgaikwad (approved):

fabianvf (approved):

PR #625, opened 2025-02-06
:bug: Fixed default params for openai

JonahSussman:
The default implementation of `_default_params` and `_get_request_payload` on `ChatOpenAI` automatically converts `max_completion_tokens` to `max_tokens`, which is undesirable for certain OpenShift AI instances.

[link](https://github.com/langchain-ai/langchain/blob/db8201d4dafb533133cc51c87c7ef011e546e03f/libs/partners/openai/langchain_openai/chat_models/base.py#L1992-L2013)

```python
@property
def _default_params(self) -> Dict[str, Any]:
    """Get the default parameters for calling OpenAI API."""
    params = super()._default_params
    if "max_tokens" in params:
        params["max_completion_tokens"] = params.pop("max_tokens")

    return params

def _get_request_payload(
    self,
    input_: LanguageModelInput,
    *,
    stop: Optional[List[str]] = None,
    **kwargs: Any,
) -> dict:
    payload = super()._get_request_payload(input_, stop=stop, **kwargs)
    # max_tokens was deprecated in favor of max_completion_tokens
    # in September 2024 release
    if "max_tokens" in payload:
        payload["max_completion_tokens"] = payload.pop("max_tokens")
    return payload
```

### DIFF
--- a/kai/llm_interfacing/model_provider.py
+++ b/kai/llm_interfacing/model_provider.py
@@ -57,6 +57,28 @@ class ModelProvider:
             case "ChatOpenAI":
                 model_class = ChatOpenAI
 
+                # NOTE(JonahSussman): This is a hack to prevent `max_tokens`
+                # from getting converted to `max_completion_tokens`
+
+                @property
+                def _default_params(self: ChatOpenAI) -> dict[str, Any]:
+                    return super(ChatOpenAI, self)._default_params
+
+                ChatOpenAI._default_params = _default_params
+
+                def _get_request_payload(
+                    self: ChatOpenAI,
+                    input_: LanguageModelInput,
+                    *,
+                    stop: list[str] | None = None,
+                    **kwargs: Any,
+                ) -> dict:
+                    return super(ChatOpenAI, self)._get_request_payload(
+                        input_, stop=stop, **kwargs
+                    )
+
+                ChatOpenAI._get_request_payload = _get_request_payload
+
                 defaults = {
                     "model": "gpt-3.5-turbo",
                     "temperature": 0.1,
--- a/kai/llm_interfacing/model_provider.py
+++ b/kai/llm_interfacing/model_provider.py
@@ -60,11 +60,11 @@ class ModelProvider:
                 # NOTE(JonahSussman): This is a hack to prevent `max_tokens`
                 # from getting converted to `max_completion_tokens`
 
-                @property
+                @property  # type: ignore[misc]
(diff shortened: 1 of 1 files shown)


fabianvf:
what does this do for the regular openai calls?

JonahSussman:
@fabianvf I tested locally and saw no difference. It looks like according to [the openai docs](https://platform.openai.com/docs/api-reference/chat/create#chat-create-max_tokens) `max_tokens` is the old param and `max_completion_tokens` is the new one. Both work for every model, aside from `o1` (and I'm assuming `o3-mini`) where you _have_ to use `max_completion_tokens`.

I suppose the more robust solution would be to solve this on the OpenShift AI itself.

fabianvf:
@JonahSussman what happens if we supply both?

kai/llm_interfacing/model_provider.py
```diff
@@ -57,6 +57,28 @@ def __init__(
...
+                # NOTE(JonahSussman): This is a hack to prevent `max_tokens`
+                # from getting converted to `max_completion_tokens`
+
+                @property  # type: ignore[misc]
+                def _default_params(self: ChatOpenAI) -> dict[str, Any]:
+                    return super(ChatOpenAI, self)._default_params
+
+                ChatOpenAI._default_params = _default_params  # type: ignore[method-assign]
+
+                def _get_request_payload(
```
pranavgaikwad, line 69: does the configureable fields still work with this? sorry i don't know the specifics of the api enough

JonahSussman:
@fabianvf Supplying both fields results in:

```
ValueError: Configuration key max_completion_tokens not found in client
```

Which makes me think that that `max_completion_tokens` doesn't actually exist, it's just substituted at runtime (which makes sense why they did it that way)... Trying to think of a smarter way to do this.

kai/llm_interfacing/model_provider.py
JonahSussman: Yes, they should AFAIK

JonahSussman:
I changed it so that for the o1 and o3 family of models, we use the original functions, and monkey patch for all others.

fabianvf:
@JonahSussman can you file a follow-up issue for a more permanent fix? I wonder if it would make the most sense to try the unmodified configuration, catch the error and monkeypatch the legacy behavior. That way we're only changing it in the case we know it's already broken. Plus that way if other models also drop it we're not breaking them

fabianvf (approved):

PR #628, opened 2025-02-06
:sparkles: Replace max_iterations with max_llm_calls

fabianvf:
- Add a counter to the model provider to optionally limit the number of requests that can be made.
- Remove all references to max_iterations (except for in the rpc_server params, we can remove that in a few releases)

### DIFF
--- a/kai/llm_interfacing/model_provider.py
+++ b/kai/llm_interfacing/model_provider.py
@@ -22,7 +22,17 @@ from kai.logging.logging import get_logger
 LOG = get_logger(__name__)
 
 
+class LLMCallBudgetReached(Exception):
+    def __init__(
+        self, message: str = "The defined LLM call budget has been reached"
+    ) -> None:
+        super().__init__(message)
+
+
 class ModelProvider:
+
+    llm_call_budget: int = -1
+
     def __init__(
         self,
         config: KaiConfigModels,
@@ -33,6 +43,7 @@ class ModelProvider:
         self.llm_retry_delay: float = config.llm_retry_delay
         self.demo_mode: bool = demo_mode
         self.cache = cache
+        self.llm_call_budget = config.llm_call_budget
 
         model_class: type[BaseChatModel]
         defaults: dict[str, Any]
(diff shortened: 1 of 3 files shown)


kai/llm_interfacing/model_provider.py
```diff
@@ -22,7 +22,17 @@
 LOG = get_logger(__name__)
 
 
+class LLMCallBudgetReached(Exception):
+    def __init__(
+        self, message: str = "The defined LLM call budget has been reached"
```
shawn-hurley, line 27: Should we say the budget that was set?

kai/llm_interfacing/model_provider.py
```diff
@@ -22,7 +22,17 @@
...
+    ) -> None:
+        super().__init__(message)
+
+
 class ModelProvider:
+
+    llm_call_budget: int = -1
```
shawn-hurley, line 34: Is this approach going to cause issues with Jonah's async PR?

fabianvf: @JonahSussman

JonahSussman: I don't think so. Each llm call can still check if the budget is reached before actually making the call. Asyncio is cooperatively multitasked

kai/rpc_server/server.py
```diff
@@ -463,7 +456,9 @@ class GetCodeplanAgentSolutionParams(BaseModel):
     file_path: Path
     incidents: list[ExtendedIncident]
 
+    # Deprecated in favor of llm_call_budget
```
JonahSussman, line 459: https://docs.pydantic.dev/latest/concepts/fields/#deprecated-fields You can make this a bonafide deprecated field if you want

kai/llm_interfacing/model_provider.py
shawn-hurley: I don't know. wouldn't multple calls to get_code_plan_solution cause the budget for ALL calls in the system to be reset (AFAICT Model provider is shared).

kai/rpc_server/server.py
JonahSussman: Or you could alias the other field to the new one

fabianvf: +1 that's a good idea

kai/llm_interfacing/model_provider.py
fabianvf: are we supporting multiple calls? The whole system kind of breaks down in that case anyway doesn't it?

shawn-hurley: It is, but it might be nice to have this not be one more thing that we have to remember if/when we do have multiple requests was my other thought.

But maybe we have to re-architect the initialization of task managers/agents at that time.

JonahSussman:
We are no longer using the Python version of Kai

PR #652, opened 2025-02-12
:sparkles: Added more notifications

JonahSussman:
- Created a `Chatter` class that you can use to send chat notifications
- Used Python's [contextvars](https://docs.python.org/3/library/contextvars.html) to create asyncio-ready solution for a global chatter

Also, did you guys know about `ExitStack()`? You basically get Go's `defer` statements, but in Python. Magic!


kai/reactive_codeplanner/agent/analyzer_fix/agent.py
```diff
@@ -203,11 +214,13 @@ def guess_language(code: str, filename: Optional[str] = None) -> str:
         all_lexers = lexers.get_all_lexers()
         largest_rv_lexer = None
         largest_rv = 0.0
+        not_found_lexers: list[str] = []
         for name, _, _, _ in all_lexers:
             try:
                 lexer_found = cast(LexerMeta, lexers.get_lexer_by_name(name))
             except ClassNotFound:
-                logger.log(DEBUG, "unable to find lexer class for name: %s", name)
+                # FIXME: This is adding significant time to our request
```
fabianvf, line 222: How significant?

kai/rpc_server/server.py
```diff
@@ -679,35 +652,41 @@ class OverallResult(TypedDict):
...
+        msg += "</details>\n"
+        msg += "<details><summary>Ignored tasks:</summary>\n"
+        msg += "<ul>\n"
+        for task in set(app.task_manager.ignored_tasks) - initial_ignored_tasks:
+            msg += f"<li>{task}</li>\n"
+        msg += "</ul>\n"
+        msg += "</details>\n"
+        chatter.get().chat_markdown(msg)
 
-        app.rcm.reset(agent_solution_snapshot)
```
fabianvf, line 697: Do we not need this anymore for some reason?

fabianvf (commented):
Few small questions but this is really nice, I like the implementation

kai/rpc_server/server.py
JonahSussman: I put the reset in the `ExitStack` above, so the rcm will always reset if an exception is encountered. `ExitStack`s sort of like Go's defer statement

fabianvf: I don't think we're worried about an exception here though, isn't this just to reset the FS state before we send it back to the IDE?

fabianvf: ah never mind I see

kai/reactive_codeplanner/agent/analyzer_fix/agent.py
JonahSussman: I put this test script together:

```python
import time

from kai.reactive_codeplanner.agent.analyzer_fix.agent import guess_language


time_before = time.time()

guess_language(
"""
import java.io.*;

public class Main {
    public static void main(String[] args) {
        System.out.println("Hello, World!");
    }
}
""",
"tester.java"
)

time_after = time.time()

print(f"Time taken: {time_after - time_before} seconds")
```

Here are the results:
```sh
(venv) [jonah@fedora kai-jonah]$ python tester.py
Time taken: 5.306198358535767 seconds
(venv) [jonah@fedora kai-jonah]$ python tester.py
Time taken: 4.978027820587158 seconds
(venv) [jonah@fedora kai-jonah]$ python tester.py
Time taken: 4.903670072555542 seconds
```

fabianvf: wow yeah that's bad

fabianvf (approved):

PR #653, opened 2025-02-13
:sparkles: Adding cache file name resolution to the task's

shawn-hurley:
* This makes it easier to determine what the path should be for each task, as tasks have more and more divergent data
* Removes the cache overwrite for SymbolNotFound and PackageDoesNotExistErrors
* Adds the depth to file tree to see effort level at work.
* In the future we should consider enhancing this, so that tasks can add data to the trace to make debugging even easier

### DIFF
--- a/kai/cache.py
+++ b/kai/cache.py
@@ -1,5 +1,4 @@
 import platform
-import re
 from abc import ABC, abstractmethod
 from pathlib import Path
 from typing import Any, Optional
@@ -12,7 +11,6 @@ from langchain_core.prompt_values import PromptValue
 from kai.jsonrpc.util import AutoAbsPath
 from kai.logging.logging import TRACE, get_logger
 from kai.reactive_codeplanner.task_manager.api import Task, ValidationError
-from kai.reactive_codeplanner.task_runner.analyzer_lsp.api import AnalyzerRuleViolation
 
 LOG = get_logger(__name__)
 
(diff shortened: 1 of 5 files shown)


shawn-hurley:
This will need a re-generation of the cache, so it will fail CI but please take a look when you get a chance to see if the base idea make sense.

kai/reactive_codeplanner/task_runner/analyzer_lsp/api.py
```diff
@@ -91,8 +92,11 @@ def background(self) -> str:
         if self.parent is not None:
             return self.oldest_ancestor().background()
         if self.children:
+            # fmt: off
+            children_messages = "\n".join(list(set(self.incident_message)))
             message = f"""You attempted to solve the following issues in the source code you are migrating:
-Issues: {"\n".join(list(set(self.incident_message)))}"""
+Issues: {children_messages}"""
```
shawn-hurley, line 98: I made this change because black on python 3.12 could not parse this.

pranavgaikwad:
@shawn-hurley CI didnt fail :(

shawn-hurley:
I think it didn't fail because all of the first shot stuff stayed the same; this change basically only changes things at depth 1 or greater. I should have realized that.

kai/cache.py
```diff
@@ -234,32 +232,18 @@ def __init__(self, task: Task, request_type: str = "llm_request"):
...
-                if not isinstance(root, ValidationError) or (
-                    isinstance(root, ValidationError) and root.file != self.task.file
-                ):
-                    stem /= self._clean_filename(self.task.file)
-            if isinstance(self.task, AnalyzerRuleViolation):
-                stem /= self.task.violation.id
-            return root_path / stem / path.name
+            root_path = root.get_cache_path(Path("."))
+            task_path = self.task.get_cache_path(Path("."))
+            print(task_path)
```
JonahSussman, line 237: Should probably be a send to a logger to avoid stdio collision stuff or removed

JonahSussman (commented):
Seems cleaner, one small nit

kai/cache.py
shawn-hurley: good catch leftover from testing, sorry

pranavgaikwad (approved):
Thank you for taking this on. This looks a lot better than before

PR #655, opened 2025-02-13
:bug: Fixing up the symbol not found and dependency agents 

### LINKED ISSUE #650: [BUG] FQDN Search agent will put the LLM in a logic cycle
The dependency_fqdn_selection agent can make the LLM go into circle logic, where it now has no way of resolving itself. The only way we get a return value is by hitting the context window. We need to fix the agent prompt to give the LLM an out.

shawn-hurley:
* Make it so that the FQDN agent doesn't cause LLM to cycle on itself
* Make it so the FQDN agent has a system prompt, that will help with
  context window issues
* Make it so that the SymbolNotFound issues go to the maven compiler.
  This is because if the dependency is missing from the project both a
symbol not found and a package does not exist are created. If just a
symbol not found, then the issue is the Class changed.

This is built on #653 to make debugging easier. 

After this change, with openshift-ai, I can fix the DatabaseMigrationStartup.java file in around 6 minutes, where before it was taking upwards of 20-30. 

fixes #650

### DIFF
--- a/kai/reactive_codeplanner/agent/dependency_agent/dependency_fqdn_selection.py
+++ b/kai/reactive_codeplanner/agent/dependency_agent/dependency_fqdn_selection.py
@@ -2,7 +2,7 @@ from dataclasses import dataclass
 from typing import Any, Optional
 
 from jinja2 import Template
-from langchain_core.messages import HumanMessage
+from langchain_core.messages import HumanMessage, SystemMessage
 
 from kai.llm_interfacing.model_provider import ModelProvider
 from kai.logging.logging import get_logger
@@ -26,6 +26,22 @@ class FQDNDependencySelectorRequest(AgentRequest):
     query: list[str]
     times: int
 
+    def from_selector_request(
+        ask: "FQDNDependencySelectorRequest", new_query: list[str]
+    ) -> "FQDNDependencySelectorRequest":
+        """Will create a new selector request bumping the times by one"""
+        req = FQDNDependencySelectorRequest(
+            ask.file_path,
+            msg=ask.msg,
+            code="",
+            query=new_query,
+            times=ask.times + 1,
+            task=ask.task,
+            background=ask.background,
+        )
+        req.cache_path_resolver = ask.cache_path_resolver
+        return req
+
 
 @dataclass
 class FQDNDependencySelectorResult(AgentResult):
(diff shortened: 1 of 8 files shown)


kai/reactive_codeplanner/task_runner/dependency/api.py
```diff
@@ -1,8 +0,0 @@
-from dataclasses import dataclass
-
-from kai.reactive_codeplanner.task_manager.api import ValidationError
-
-
-@dataclass
-class DependencyValidationError(ValidationError):
```
shawn-hurley, line 7: @fabianvf I don't see this used anywhere so I removed if I could get a second pair of eyes that would be great.

kai/reactive_codeplanner/agent/dependency_agent/dependency_fqdn_selection.py
```diff
@@ -26,6 +26,22 @@ class FQDNDependencySelectorRequest(AgentRequest):
     query: list[str]
     times: int
 
+    def from_selector_request(
```
shawn-hurley, line 29: This is needed to pass cache_path_resolver to the next run of execute.

kai/reactive_codeplanner/agent/dependency_agent/dependency_fqdn_selection.py
```diff
@@ -42,22 +58,22 @@ class __llm_response:
     def __init__(self, model_provider: ModelProvider) -> None:
         self._model_provider = model_provider
 
-    message = Template(
+    system_tmpl = Template(
```
shawn-hurley, line 61: Using a system template makes the context window smaller and the LLM behave better.

kai/reactive_codeplanner/agent/maven_compiler_fix/agent.py
```diff
@@ -21,8 +21,6 @@ class MavenCompilerAgent(Agent):
         """{{ background }}
     I will give you compiler errors and the offending line of code, and you will need to use the file to determine how to fix them. You should only use compiler errors to determine what to fix.
 
-    Make sure that the references to any changed types are kept.
```
shawn-hurley, line 24: This allows the LLM to revert certain things and, in my testing, gets slightly better results for specific situations. 

It is imperfect, and if we need to add it back, we can.

JonahSussman (approved):
LGTM. Not specific to this PR, but there's some funkiness happening with how we handle `Task`s and `Agent`s

kai/reactive_codeplanner/task_runner/dependency/task_runner.py
```diff
@@ -36,11 +32,7 @@ class DependencyTaskResponse:
 class DependencyTaskRunner(TaskRunner):
     """TODO: Add Class Documentation"""
 
-    handled_type = (
-        DependencyValidationError,
-        SymbolNotFoundError,
-        PackageDoesNotExistError,
-    )
+    handled_type = (PackageDoesNotExistError,)
```
pranavgaikwad, line 35: What happens when a symbolnotfound is actually caused because of a missing dep?

shawn-hurley: A package not found error is also present and happens before that. There are then two scenarios because packages not found are handled first. 

1. We added the package, and this issue is resolved indirectly
2. We were unable to add the package. This could be because it is a bad update (see the Jakarta.sql) example, in the case we need to back out the specific change, in other cases, it may be that the name of the thing has changed or something. In this case, it is not a package error; it is an error in that file that needs to be fixed.

PR #667, opened 2025-02-18
:book: Adding demo mode documentation.

shawn-hurley:
* Adding the reason for using demo mode
* The parts that are needed for demo mode
* How to configure them
* Warning around ordering of solving incidents to make sure cache hits still happen


docs/demo_mode.md
```diff
@@ -0,0 +1,53 @@
+# Demo Mode
+
+There may be times when you have to give a demo of KaiKai in a restricted network setting. This could be because you don't have access to LLMs, the network will be too slow to contact them, or you don't want to wait for the LLM while showcasing what Kai can do for you!
```
JonahSussman, line 3:
```suggestion
There may be times when you have to give a demo of Kai in a restricted network setting. This could be because you don't have access to LLMs, the network will be too slow to contact them, or you don't want to wait for the LLM while showcasing what Kai can do for you!
```

JonahSussman, line 3:
```suggestion
There may be times when you have to give a demo of Kai in a restricted network setting. This could be because you don't have access to LLMs, the network will be too slow to contact them, or you don't want to wait for the LLM while showcasing what Kai can do for you!
```

JonahSussman (changes requested):

JonahSussman (approved):
LGTM

PR #675, opened 2025-02-20
:sparkles: Added continuation support for Amazon Bedrock

JonahSussman:
Closes #136 
Closes #350

I refactored `model_provider.py`, putting each `ModelProvider` into its own class. This will allow you to override certain model-specific functionality. For example, a custom LLM invoke for Bedrock (the main reason I refactored)

Bedrock will now continue to generate tokens even if `max_tokens` is reached. Look at the following Python script and attached logs:

```python
from langchain.globals import set_debug, set_verbose

from kai.kai_config import KaiConfigModels, SupportedModelProviders
from kai.llm_interfacing.model_provider import ModelProvider

set_verbose(True)
set_debug(True)

m = ModelProvider.from_config(
    config=KaiConfigModels(
        provider=SupportedModelProviders.CHAT_BEDROCK,
        args={"model_id": "us.anthropic.claude-3-5-sonnet-20241022-v2:0"},
    )
)

print(m.invoke("Generate a long poem. 10,000 words. It must be very long.").content)
```

[tester.log](https://github.com/user-attachments/files/18881971/tester.log)

Ugly langchain color codes aside, you can see how we do two requests now, resulting in one continuous poem at the end.

I tested locally with Amazon Bedrock and OpenAI, but I would like some assistance testing other providers to make sure I didn't mess anything else up.

Separately, we should probably think about how to integrate streaming into the project. It should be fairly straightforward now with the `Chatter` class. We can create a message with a separate `messageToken` and update it as we get more chunks. WDYT?

PR #684, opened 2025-02-24
:sparkles: Added Anthropic and Groq support

devjpt23:
Once this PR is merged, Anthropic and Groq models can be accessed via the IDE by changing the model in `provider-settings.yaml`. Below shows the way the file needs to be changed.
```

environment:
  ALWAYS_APPLIED_KEY: "envvar to be set regardless of which model is active"

# This is a collection of model configurations that may be used.  The `&active`
# anchor is used to reference the model to the active node below.  The extension will
# use the active node for configuration the kai-rpc-server.
models:
  ChatAnthropic:
    environment:
      ANTHROPIC_API_KEY: "YOUR_ANTHROPIC_API_KEY"

    provider: "ChatAnthropic"
    args:
      model: "claude-3-5-sonnet-20241022"

  
  ChatGroq: &active
    environment:
      GROQ_API_KEY: "YOUR_GROQ_API_KEY

    provider: "ChatGroq"
    args:
      model: "mixtral-8x7b-32768"
```


jwmatthews:
@devjpt23 we will likely want to bring in https://github.com/konveyor/kai/pull/675 first and then your PR after it’s ready and tested.

It would be worth you taking a look at #675 and being familiar, we think the continuation support may be the fix to problems we’ve seen working with Claude and other models via Amazon bedrock in past.

PR #685, opened 2025-02-24
:bug: Removed monkey patch, OpenShift AI now works with `max_completion_tokens`

### LINKED ISSUE #627: Monkey patch `_default_params` and `_get_request_payload` only if needed
Following up on https://github.com/konveyor/kai/pull/625, we should change it so that we try the catch the error and monkey patch the legacy behavior. That way we're only changing it in the case we know it's already broken. Plus that way if other models also drop it we're not breaking them.

shawn-hurley:
fixes #627


shawn-hurley:
removed all the code to not do the validate_environment. Not all models use pydantic model validation.

pranavgaikwad (approved):

JonahSussman (approved):
ACK

PR #686, opened 2025-02-24
:bug: Added back the ability to get tokens for input and output

### LINKED ISSUE #373: Capture number of tokens in a request and response when possible
We've run into a few situations where it would benefit us if we had a better view of the number of tokens consumed in a request and response.  

Let's augment the data we are capturing for [tracing](https://github.com/konveyor/kai/blob/main/docs/contrib/tracing.md) and add in any extra info we may get back from the LLM via 'response_metadata'.

Current understanding is that for some models, the response includes metadata that breaks out the number of tokens used in the request and the response.

### LINKED ISSUE #617: Estimation of usage for tokens sent and/or consumed
Its not easy for a user to understand how much tokens are sent or consumed.

This can have potential affects. 
- Billing; where e.g. estimating the cost on a provider like OpenAI
- Sizing/power/cost etc. in cases of Privately hosted models using in house or cloud GPUs. 
- General code/app estimations: For someone doing this at a larger scale on multiple projects, this could be very useful. Or for a particular incident type that might affect multiple projects or files. 

A metric around the estimated amount of tokens passed and consumed would be very helpful as a start. 
There might be other mechanisms to tackle this to.

shawn-hurley:
fixes #617 
fixes #373


kai/llm_interfacing/callback/token_output_callback.py
```diff
@@ -0,0 +1,56 @@
...
+    def on_chat_model_start(
+        self,
+        serialized: dict[str, Any],
+        messages: list[list[BaseMessage]],
+        **kwargs: Any
+    ) -> None:
+        flat_messages = [item for sublist in messages for item in sublist]
+        try:
+            tokens = self.model.get_num_tokens_from_messages(flat_messages)
+            LOG.info("input tokens: %s", tokens)
```
pranavgaikwad, line 50: no f-strings? I didn't know you could do this in py

shawn-hurley: AFAIK that the log.info works like this

kai/llm_interfacing/callback/token_output_callback.py
```diff
@@ -0,0 +1,56 @@
...
+        except Exception:
+            # Here we fall back to a default encoding if no model is found.
+            enc = tiktoken.get_encoding("cl100k_base")
+            tokens = len(enc.encode(get_buffer_string(flat_messages)))
+            LOG.info("input tokens: %s", tokens)
+        return
```
JonahSussman, line 56: Extra return?

```suggestion
```

JonahSussman (approved):
LGTM just one nit

Issue #665, opened 2025-02-17
[bug] Investigate errors in equality checking


fabianvf:
Multiple users have reported issues being detected as children of themselves and retried indefinitely. This is likely an error in the equality checking, potentially stemming from our combination of multiple incidents into a single task. We have not yet identified a clear reproducer or root cause, so if anyone has seen this recently, please comment and upload your logs so we can investigate further.

shawn-hurley:
I believe that this is closed with the merge of #692 if folks are still seeing this issue, please reopen it!

PR #692, opened 2025-02-25
:bug: CollapsableErrors assume validation field values are empty

### LINKED ISSUE #677: The kai-rpc-server log file has grown by more than 2GB in 10 minutes while resolving a single issue
While resolving the issue `@MessageDriven - EJBs are not supported in Quarkus` for `src\main\java\com\redhat\coolstore\service\OrderServiceMDB.java` the log file grew over 2.5GB in less than 15 minutes, and the issue was still unresolved

[kai-analyzer-server.log](https://github.com/user-attachments/files/18892157/kai-analyzer-server.log)
[6-Konveyor-Analyzer.log](https://github.com/user-attachments/files/18892161/6-Konveyor-Analyzer.log)

Here is a fragment of the log file, in which I included the beginning and part of the messages that are repeated hundreds of times

[partial-rpc-server.log](https://github.com/user-attachments/files/18892274/partial-rpc-server.log)

**Env:**

- Windows 11 Pro
- Extension v0.0.11 with maximum effort set

shawn-hurley:
* When doing comparison for collapsable errors, we assume that only the fields that matter are different, and everything else is reset.
* We were not reseting the column for collapsable errors, causing them to look different for both fuzzy match and equality.
* When it was unable to be solved, due to an error, we infinitly added parent task as a child task.

fixes #677

### DIFF
--- a/kai/reactive_codeplanner/task_runner/compiler/maven_validator.py
+++ b/kai/reactive_codeplanner/task_runner/compiler/maven_validator.py
@@ -149,6 +149,10 @@ class CollapsedMavenCompilerError(ABC, MavenCompilerError):
     def get_collapsable_key(self) -> str:
         pass
 
+    @abstractmethod
+    def remove_unused_fields(self) -> None:
+        pass
+
 
 # Subclasses for specific error categories
 @dataclass(eq=False)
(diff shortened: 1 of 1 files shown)


pranavgaikwad:
This might also be the fix we are looking for in #665

shawn-hurley:
I think it is TBH but didn't know if there was another one, I didn't look into the analyzer types yet to verify.

pranavgaikwad (approved):

JonahSussman (approved):

PR #696, opened 2025-02-26
:sparkles: Handle retries and ignored tasks by not adding children

shawn-hurley:
* This will short circut the handling of a task when it is ignored to not add children
* Also added the logic to on moving a task to ignored, we revert the repo back to the state it was in when the task started.
* Had to remove the spawning result history from RepoContextSnapshot because of cycle imports of task -> RCS -> SpawningResult -> ReflectionTask -> task.

### DIFF
--- a/kai/reactive_codeplanner/task_manager/task_manager.py
+++ b/kai/reactive_codeplanner/task_manager/task_manager.py
@@ -401,7 +401,7 @@ class TaskManager:
             self.rcm.reset(task._snapshot_before_work)
             chatter.get().chat_markdown(
                 f"Task {task.__class__.__name__} was not resolved. resetting repo state to before task was tried)"
-                f"<details><summary>Details</summary>\n{resolved_task.markdown()}</details>\n"
+                f"<details><summary>Details</summary>\n{task.markdown()}</details>\n"
             )
 
     def stop(self) -> None:
--- a/kai/reactive_codeplanner/task_manager/task_manager.py
+++ b/kai/reactive_codeplanner/task_manager/task_manager.py
@@ -222,6 +222,8 @@ class TaskManager:
                 continue
 
             logger.info("Yielding task: %s", task)
+            if not task._snapshot_before_work:
+                task._snapshot_before_work = self.rcm.snapshot
             yield task
             # If our depth is 0, we won't follow up on issues anyway
             # We do lose the ability to verify a solution worked, so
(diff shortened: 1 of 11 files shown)


kai/reactive_codeplanner/task_manager/task_manager.py
```diff
@@ -295,6 +297,10 @@ def handle_new_tasks_after_processing(
             for t in similar_tasks:
                 unprocessed_new_tasks.remove(t)
             self.handle_ignored_task(task)
+            # Once we have a retry or an ignored task, we should wait to add
+            # children until the task is completed.
+            # On ignored task, we now revert to the before snapshot work
+            return
```
pranavgaikwad, line 303: just trying to understand, would this also stop yielding other tasks within the same depth level?

shawn-hurley: no because this is in the handle_new_tasks_after_processing method, it stops doing more work there. The only work that is below this, is the attaching child tasks, which we don't want to do. The task_queue either gets the retried task back on it, or it is moved to ignored and the next task at the same depth will be taken.

fabianvf (approved):

kai/reactive_codeplanner/task_manager/task_manager.py
```diff
@@ -382,6 +388,11 @@ def handle_ignored_task(self, task: Task) -> None:
             chatter.get().chat_simple(
                 f"Task {task} was not resolved. Ignoring... ({task.retry_count}/{task.max_retries})"
             )
+            logger.info("ignoring task, reverting to pre-task snapshot")
+            self.rcm.reset(task._snapshot_before_work)
+            chatter.get().chat_simple(
```
fabianvf, line 393:
```suggestion
            chatter.get().chat_markdown(
```

kai/reactive_codeplanner/task_manager/task_manager.py
```diff
@@ -382,6 +388,11 @@ def handle_ignored_task(self, task: Task) -> None:
...
+                f"Task {task} was not resolved. resetting repo state to before task was tried)"
```
fabianvf, line 394:
```suggestion
                f"Task {task.__class__.__name__} was not resolved. resetting repo state to before task was tried)"
                f"<details><summary>Details</summary>\n{resolved_task.markdown()}</details>\n"
```

fabianvf (approved):

PR #723, opened 2025-03-06
:bug: Handle filtered incidents in the and/or conditions

shawn-hurley:


JonahSussman (approved):
ACK

pranavgaikwad (approved):

savitharaghunathan:
tested the workflow with the binaries from the build pipeline and it works. Once the custom rule violation is fixed, it doesnt retrigger

PR #731, opened 2025-03-07
:bug: Added label configuration to always be updated by analysis with labels

shawn-hurley:
* Adding partial analaysis call at start of get_code_plan
* TODO: old child tasks are marked as resolved on subsequent runs
* TODO: need to manage the unprocess_files to remove them on seed_tasks


kai/reactive_codeplanner/agent/dependency_agent/dependency_agent.py
```diff
@@ -300,6 +300,12 @@ def execute(self, ask: AgentRequest) -> AgentResult:
         # We are going to give back the response, The caller should be responsible for running the code generated by the AI.
         # If we have not take the actions step wise, in the LLM, we need to run all but editor here
         # and give that information to the caller.
+        if (
+            not llm_response.final_answer
+            and maven_search is not None
+            and maven_search.version
+        ):
+            llm_response.final_answer = "found dependency"
```
shawn-hurley, line 308: I think this may fix some of the dep issues, while testing before search went down, I notice that we get the right dep, but no final answer. For now, if we do get a dep and no final answer, (version is only found when a valid search) then we should act like we found the solution.

pranavgaikwad (approved):

fabianvf (approved):
LGTM, still need to test

kai/analyzer.py
```diff
@@ -109,18 +113,27 @@ def __init__(
...
+        if label_selector is None and self.labels is not None:
+            label_selector = self.labels
+
+        if incident_selector is not None:
+            self.incident_selector = incident_selector
+
         request_params: dict[str, Any] = {
-            "label_selector": label_selector,
-            "incident_selector": incident_selector,
+            "label_selector": self.labels,
```
shawn-hurley, line 133: bug, this needs to move to label_selector.

kai/reactive_codeplanner/task_manager/task_manager.py
```diff
@@ -68,19 +68,28 @@ def set_seed_tasks(self, *tasks: Task) -> None:
         # to resolve any issues found even if they've been seen before
         self.processed_tasks = set()
         self.ignored_tasks = list()
+        self.processed_files = list()
```
shawn-hurley, line 71: bug needs to be unprocessed files as well.

kai/analyzer.py
```diff
@@ -109,18 +113,27 @@ def __init__(
...
+
+        if label_selector is not None and self.labels is None:
+            self.labels = label_selector
+
+        if label_selector is None and self.labels is not None:
+            label_selector = self.labels
+
+        if incident_selector is not None:
+            self.incident_selector = incident_selector
+
```
JonahSussman, line 131: Nit: Do we want to statefully modify the class here or should we use the class as the default and have the params as modifications?

JonahSussman (approved):
One nit, otherwise LGTM

kai/analyzer.py
shawn-hurley: The idea is that if they are not set on startup, but set after, and then you run analysis, we capture them and save them for later (specifically in the case that you then run code plan).

PR #732, opened 2025-03-07
:bug: set label selector via initialize

pranavgaikwad:


shawn-hurley:
@pranavgaikwad I added this commit to my PR in #731 because I added some other handling in the analysis call with labels as well, just in case.

Issue #722, opened 2025-03-06
Maven parsing error (see description for details)


pranavgaikwad:
Seeing following error in code plan output:

```
ERROR - 2025-03-06 21:22:59,897 - kai.kai.reactive_codeplanner.task_manager.task_manager - Thread-1 - [task_manager.py:99 - execute_task()] - Unhandled exception executing task OtherError<loc=/Users/pgaikwad/Projects/coolstore/target:-1:-1, message=Unknown error occurred during Maven build.>(priority=6(0), depth=1, retries=0)
Traceback (most recent call last):
  File "kai/reactive_codeplanner/task_manager/task_manager.py", line 97, in execute_task
  File "contextlib.py", line 81, in inner
  File "kai/reactive_codeplanner/task_runner/compiler/compiler_task_runner.py", line 83, in execute_task
IsADirectoryError: [Errno 21] Is a directory: '/Users/pgaikwad/Projects/coolstore/target'
```

This happens because something ends up being classified as `OtherError` and the file path for that error is set to target dir. Here's the maven output:

```
[INFO] Scanning for projects...
[INFO] 
[INFO] -------------------< com.redhat.coolstore:coolstore >-------------------
[INFO] Building coolstore-quarkus 1.0.0-SNAPSHOT
[INFO]   from pom.xml
[INFO] --------------------------------[ jar ]---------------------------------
[INFO] 
[INFO] --- clean:3.2.0:clean (default-clean) @ coolstore ---
[INFO] Deleting /Users/pgaikwad/Projects/coolstore/target
[INFO] 
[INFO] --- resources:3.3.1:resources (default-resources) @ coolstore ---
[INFO] Copying 2516 resources from src/main/resources to target/classes
[INFO] 
[INFO] --- quarkus:3.12.3:generate-code (default) @ coolstore ---
[INFO] 
[INFO] --- compiler:3.13.0:compile (default-compile) @ coolstore ---
[INFO] Recompiling the module because of changed source code.
[INFO] Compiling 25 source files with javac [debug release 21] to target/classes
[INFO] ------------------------------------------------------------------------
[INFO] BUILD FAILURE
[INFO] ------------------------------------------------------------------------
[INFO] Total time:  3.117 s
[INFO] Finished at: 2025-03-06T21:25:37+05:30
[INFO] ------------------------------------------------------------------------
[...]

PR #733, opened 2025-03-08
Update task_manager.py

RohanMishra315:
fixes: #722
1. Check you have corrected Java version
2. IsADirectoryError separately handeled.

### DIFF
--- a/kai/reactive_codeplanner/task_manager/task_manager.py
+++ b/kai/reactive_codeplanner/task_manager/task_manager.py
@@ -87,24 +87,46 @@ class TaskManager:
             self.priority_queue.push(task)
             logger.info("Seed task %s added to stack.", task)
 
-    def execute_task(self, task: Task) -> TaskResult:
-        logger.info("Executing task: %s", task)
-        agent = self.get_agent_for_task(task)
-        logger.info("Agent selected for task: %s", agent)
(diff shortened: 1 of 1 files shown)


pranavgaikwad (changes requested):
Thank you for attempting a fix for this. Really appreciate it. A few suggestions...It would be better to fix this issue at the root instead of catching it as an exception. What is happening is that something that is not an error is being parsed as an OtherError with a wrong error. The correct place to fix this would be in file https://github.com/konveyor/kai/blob/main/kai/reactive_codeplanner/task_runner/compiler/maven_validator.py

This is where we run the maven command and parse its output. To understand how we are parsing the errors, take a look at this function https://github.com/konveyor/kai/blob/725cd68ea1a868d5c9f670cbc70653ce37ab228f/kai/reactive_codeplanner/task_runner/compiler/maven_validator.py#L594-L623

Most probably the fix will be in this function (you may have to reproduce the issue to figure that out). To reproduce, you can simply create a new test in file https://github.com/konveyor/kai/blob/main/tests/test_maven_validator.py

Use the output in the issue description. Take a look at existing tests to figure out how the `parse_error_lines` function is called. And finally, reach out if you need help :)

Thank you again.

PR #739, opened 2025-03-13
:ghost: validated usage of custom wrapper for tools for models that don't have a tool api

pranavgaikwad:
See preview here https://github.com/pranavgaikwad/kai/blob/toolUsage/notebooks/followup_agent/00_do_followup_work.ipynb

Scroll all the way down to `Using tools with models that do not support tool calling` section


fabianvf (approved):
This is cool

PR #754, opened 2025-03-25
:bug: add default cache prime step

### LINKED ISSUE #694: Generating resolutions for issues from a stale analysis report breaks things
This is an interesting case where if you have an analysis report from prior runs, and you start the server but forget to re-run analysis, cache doesn't get primed and things break.

savitharaghunathan:
fixes #694 

@pranavgaikwad @shawn-hurley PTAL and let me know if this the right spot to fix the issue. I am a work in progress on understanding the codebase. Any thoughts/suggestions will be appreciated. Thanks

### DIFF
--- a/kai/rpc_server/server.py
+++ b/kai/rpc_server/server.py
@@ -282,6 +282,15 @@ def initialize(
             ],
         )
 
+        app.log.info("Priming analyzer cache with default run")
+
+        app.analyzer.run_analyzer_lsp(
+            label_selector=app.config.analyzer_lsp_label_selector or "",
+            included_paths=[],
+            incident_selector=None,
+            scoped_paths=None,
+            reset_cache=False,
+        )
     except Exception:
         server.send_response(
             id=id,


kai/rpc_server/server.py
```diff
@@ -282,6 +282,15 @@ def initialize(
             ],
         )
 
+        app.log.info("Priming analyzer cache with default run")
```
pranavgaikwad, line 285: can the messaging be "Running analysis" instead?

pranavgaikwad (changes requested):

kai/rpc_server/server.py
savitharaghunathan: @pranavgaikwad you mean change `Priming analyzer cache with default run` -> `Running analysis` ?

pranavgaikwad: yeah

kai/rpc_server/server.py
```diff
@@ -282,6 +282,15 @@ def initialize(
...
+
+        app.analyzer.run_analyzer_lsp(
```
shawn-hurley, line 287: I wonder if this can be fired and returned so we are ready to work? 

I guess the question is, how much time does this add to start up, and do we want to "do" something different

Issue #753, opened 2025-03-24
Scenario for springboot2 -> 3 migration focussing on authentication


savitharaghunathan:
Create a migration scenario that demonstrates how to upgrade a Spring Boot 2.x application to Spring Boot 3.x, specifically focusing on changes to authentication and security configuration.

Goals:
* Migrate a simple login application from Spring Boot 2.7.x to 3.2.x
* Update security configuration due to deprecation of WebSecurityConfigurerAdapter
* Replace javax.* imports with jakarta.*
* Ensure compatibility with Java 17+

Additional info:
* The old configure(HttpSecurity) approach is no longer valid in Spring Security 6+
* Jakarta EE namespace changes are mandatory

Associated tasks

- [x] Research and identify the scenario
- [x] Create an app with the identified scenario
- [x] Create a migrated version 
- [x] Identify custom rules and create them
- [x] run the app with custom rules in kai
- [x] record any changes thats needed, contribute back to custom rules
- [x] Create a scenario doc and share it with QE

savitharaghunathan:
gh repo - https://github.com/savitharaghunathan/springboot_simplelogin

savitharaghunathan:
Migrated branch - https://github.com/savitharaghunathan/springboot_simplelogin/tree/springboot3

Custom rules WIP - https://github.com/savitharaghunathan/custom_rules/tree/savitharaghunathan-patch-1/springlogin

PR #762, opened 2025-04-10
:sparkles: Add scenario - springboot 2 -> 3

savitharaghunathan:
Fixes https://github.com/konveyor/kai/issues/753


docs/scenarios/springboot_2_to_3_security/README.md
```diff
@@ -0,0 +1,592 @@
...
+
+Follow the [official guide](https://github.com/konveyor/kai/blob/main/docs/installation.md) for installation instructions.
+
+## Step 2: Understand the Custom Rules
+
+Before we run the analysis, let’s understand the custom rules we are going to use for migrating from Spring Boot 2 to Spring Boot 3, focusing specifically on the removal of `WebSecurityConfigurerAdapter` and discouraged configuration overrides.
+
+### 2.1. Custom Rule for Deprecated configure(AuthenticationManagerBuilder) Method
+
+The `SecurityConfig.java` class configures users and roles using the now-discouraged configure `(AuthenticationManagerBuilder auth)` method:
```
jwmatthews, line 58: small nit, to include 'configure' method name in the syntax highlight

change:
configure `(AuthenticationManagerBuilder auth)` method:

to:
`configure(AuthenticationManagerBuilder auth)` method:

docs/scenarios/springboot_2_to_3_security/README.md
```diff
@@ -0,0 +1,592 @@
...
+- Spring Boot version: `2.7.4` -> `3.0.0`
+- Java version: `8` -> `17`
+
+![pom.xml Merge View](./images/pom_merge.png)
+
+Click on `check` icon to accept and apply the resolution.
+
+![Apply pom.xml Resolution](./images/pom_apply.png)
+
+The analysis will re-run and update the incidents.
```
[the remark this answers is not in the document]
savitharaghunathan: I have occasionally noticed that happen but not been able to reproduce all the time. Are you able to reproduce it always?

docs/scenarios/springboot_2_to_3_security/README.md
```diff
@@ -0,0 +1,592 @@
...
+    java.referenced:
+      location: METHOD
+      pattern: configure(org.springframework.security.config.annotation.authentication.builders.AuthenticationManagerBuilder) *
+  message: |
+    Overriding `configure(AuthenticationManagerBuilder)` is **discouraged** in Spring Security 5.7.
+    Instead, define a `UserDetailsService` and `PasswordEncoder` bean.
+
+    **Fix:** Move user setup into a `UserDetailsService` bean like this:
+
+    ```java
```
jwmatthews, line 119: We should consider if we are supplying too much of the exact solution in the hint.

What I mean is ideally we supply information in the hint so the LLM can adapt the answer for different related situations.  For the example we look at in this scenario I don't have a good feel if the LLM can adapt the info or if it's more using exactly what we placed in the hint.

I don't think we need to make a change right now, I'm sharing more for information purposes so we can be thinking about this with further examples.

savitharaghunathan: thats valuable feedback. I can play with the rule and see if I can update this to be generic. When I think of custom rules, I think it is highly specific because an organization might have their own lib and want to update it in a specific way. That was the motive behind having the highly specific code block here. Just sharing the perspective I had.

jwmatthews (approved):
Confirming I ran through and tested this scenario.

Worked as expected for me and able to deploy app afterwards.
I had one hiccup with analysis not refreshing, but I don't think that is related to the scenario and more a hiccup elsewhere.

sshaaf:
LGTM. thanks!

PR #773, opened 2025-05-21
:sparkles: Feature/kai solution server draft

JonahSussman:


fabianvf:
Could you add a method to tests/mcp_client.py to test the new endpoint?

PR #810, opened 2025-06-18
:bug: Make directories root group writeable for restricted environments

fabianvf:


jmontleon (approved):
ACK, tested and working

PR #815, opened 2025-06-25
Fix Azure OpenAI configuration documents

feiskyer:
Add required paramters to the doc.


feiskyer:
Thanks for reviewing. @jwmatthews  @pranavgaikwad  let's get this in?

PR #937, opened 2026-07-09
:bug: Build solution-server image from the lockfile, not pyproject floors

### LINKED ISSUE #934: Unbounded fastmcp constraint: image rebuilds pick up FastMCP 3.4.3 host-origin protection, kai-api returns 421 for all proxied requests
## Summary

Since July 8, the kai-api MCP server (`quay.io/konveyor/kai-solution-server:latest`) returns **HTTP 421 Misdirected Request** for every request when deployed behind a proxy/ingress (minikube, and likely OpenShift routes). The server is completely unreachable through the ingress.

## Root cause

Two compounding problems:

1. **The image build ignores the lockfile.** [`tools/deploy/Containerfile`](https://github.com/konveyor/kai/blob/main/kai_mcp_solution_server/tools/deploy/Containerfile) runs `pip3.12 install --no-cache-dir .` against `pyproject.toml` only — `requirements.txt` / `uv.lock` (which pin `fastmcp==2.14.3`, `mcp==1.25.0`) are never copied into the image. pyproject declares only a floor: `fastmcp>=2.8.0`, so every rebuild resolves the newest fastmcp from PyPI.
2. **FastMCP 3.4.3 enabled host/origin (DNS-rebinding) protection by default** — `http_host_origin_protection: bool = True` with `http_allowed_hosts: None`, i.e. an effectively empty allowlist. The security middleware validates the `Host` header and returns `421 "Invalid Host header"` on mismatch. Same failure mode hit other projects the same week: [modelcontextprotocol/python-sdk#1798](https://github.com/modelcontextprotocol/python-sdk/issues/1798), [CrowdStrike/falcon-mcp#291](https://github.com/CrowdStrike/falcon-mcp/issues/291), [blockscout/mcp-server#333](https://github.com/blockscout/mcp-server/issues/333), [write-up](https://dev.to/toyama0919/how-an-unbounded-fastmcp-version-constraint-took-down-production-with-421-misdirected-request-1mh1).

The merge of #932 triggered the fresh `:latest` build that picked up fastmcp ≥3.4.3 — but the dependency bumps in #932 are themselves **incidental**; any merge would have produced the same broken image. (Earlier drafts of this issue blamed the starlette bump; that was wrong — starlette has no 421 code path.)

Behind a proxy, the forwarded `Host` header is the external host/IP (e.g. `192.168.49.2` via tackle-hub's `/hub/services/kai/api` route), which never matches the empty/localhost allowlist → every request is rejected before reaching the MCP session layer.

## Evidence

kai-api pod logs — the server itself returns 421, not nginx:

```
INFO:     Starting MCP server 'KaiSolutionServer' with transport 'streamable-http' on http://0.0.0.0:8000/api
INFO:     10.244.0.16:45922 - "POST /api HTTP/1.1" 421 Misdirected Request
INFO:     10.244.0.16:37090 - "POST /api HTTP/1.1" 421 Misdirected Request
... (every request, deterministic)
```

## Impact

- All `editor-extensions` CI infrastructure tests (`@requires-minikube`) are broken
- Any deployment behind an ingress/route that preserves the external Host header is unreachable

## Reproduction

1. Deploy kai-api via the konveyor operator on minikube with ingress enabled
2. `curl -k -X POST https://$(minikube ip)/hub/services/kai/api -H "Content-Type: application/json" -d '{"jsonrpc":"2.0","method":"initialize","id":0,"params":{"protocolVersion":"2025-03-26","capabilities":{},"clientInfo":{"name":"test","version":"0.0.0"}}}'`
3. Observe HTTP 421

## Suggested fixes (increasing durability)

- **Immediate mitigation** for existing deployments: set `FASTMCP_HTTP_HOST_ORIGIN_PROTECTION=false` on the kai-api container
- **Make the image reproducible**: install from the lock (`uv sync --frozen`, or copy `requirements.txt` into the image and `pip install -r`) instead of resolving floors at build time
- **Cap the constraint** (`fastmcp>=2.8.0,<3`) until a deliberate 3.x migration, and/or configure `http_allowed_hosts` explicitly for proxied deployments

ibolton336:
## Summary

Fixes #934. The `kai-solution-server` image drifted onto **fastmcp 3.4.3**, whose `HostOriginGuardMiddleware` is enabled by default and returns `421 Misdirected Request` for every request whose `Host` header doesn't match — i.e. every request behind an ingress/route. This broke `editor-extensions` `@requires-minikube` CI and affects any proxied deployment.

### Root cause

[`kai_mcp_solution_server/tools/deploy/Containerfile`](https://github.com/konveyor/kai/blob/main/kai_mcp_solution_server/tools/deploy/Containerfile) installed the app with `pip install .`, which **resolves the `pyproject.toml` dependency floors at build time** (e.g. `fastmcp>=2.8.0`) and ignores the committed lock. `requirements.txt` / `uv.lock` pin `fastmcp==2.14.3`, `mcp==1.25.0`, `starlette==1.3.1`, but those pins never reached the image. So each `:latest` rebuild silently picked up the newest resolvable `fastmcp` — 3.4.3 once it published.

For reference, fastmcp **3.4.4** already flipped `http_host_origin_protection` back to `False` by default, but relying on that is just more floating-latest roulette; the image should install what the project locks.

### Fix

Install the pinned set from `requirements.txt` first, then the package itself with `--no-deps`:

```dockerfile
COPY requirements.txt .
...
RUN pip3.12 install --no-cache-dir -r requirements.txt \
 && pip3.12 install --no-cache-dir --no-deps .
```

This makes the image reproducible and match what the project actually tests against, and fixes the drift for **all** dependencies, not just fastmcp.

### Validation

I couldn't build the UBI image in this environment, but I verified the mechanical precondition for `--no-deps`: **every one of the 34 runtime dependencies declared in `pyproject.toml` is present in `requirements.txt`** (192 pinned packages), so `pip install --no-deps .` installs cleanly with nothing missing. A maintainer build + push of `:latest` is the final confirmation.

### Alternatives considered

- Capping `fastmcp>=2.8.0,<3` in `pyproject.toml` — smaller diff, but only patches this one dependency and leaves the image resolving every other floor at build time. Installing from the lock addresses the class of bug.

### DIFF
--- a/kai_mcp_solution_server/tools/deploy/Containerfile
+++ b/kai_mcp_solution_server/tools/deploy/Containerfile
@@ -11,10 +11,17 @@ RUN dnf --refresh -y update \
 # Copy project files
 COPY src/ src/
 COPY pyproject.toml .
+COPY requirements.txt .
 COPY README.md .
 
-# Install Python dependencies
-RUN pip3.12 install --no-cache-dir .
+# Install Python dependencies from the pinned lockfile export, then the
+# package itself without re-resolving. Installing `.` alone resolves the
+# pyproject floors at build time, which let fastmcp drift onto 3.4.3 —
+# whose Host/Origin guard 421s every request behind an ingress proxy.
+# Pinning to requirements.txt (fastmcp==2.14.3, mcp==1.25.0, ...) keeps the
+# image reproducible and matches what the project actually tests against.
+RUN pip3.12 install --no-cache-dir -r requirements.txt \
+ && pip3.12 install --no-cache-dir --no-deps .
 
 # Create a non-root user
 RUN useradd -m mcp


ibolton336:
Closing in favor of #935 (@fabianvf), which lands the same fix and more — it pins `fastmcp<3` **and** builds the image from the lockfile, plus adds a `test_host_header.py` regression test and a dedicated image-test CI workflow so this can't silently regress again. Thanks — same diagnosis, better coverage.

